PR #3, opened 2015-02-05
virtualjaguar remove unused eclass, -egl, src_configure return and install desktop icon+file.

damex:
short description of the changes:
1. eutils is not used here. at'all.
2. with egl enabled virtualjaguar does not provide any graphics output.
3. it does repeat eqmake\* on src_configure and fails to build after.
4. it does not install bundled desktop icon with desktop file.

PR #4, opened 2015-02-06
-gl, src_configure breaks building and install desktop icon&file

damex:
previously #3 
single commit pull request.


mgorny:
Sad github is not smart enough to --find-copies-harder :).

games-emulation/virtualjaguar/virtualjaguar-2.1.2-r1.ebuild
```diff
@@ -0,0 +1,58 @@
...
+	dev-qt/qtgui:4[-egl]
+	dev-qt/qtopengl:4[-egl]
+	dev-libs/libcdio"
+DEPEND="${RDEPEND}
+	>=sys-devel/gcc-4.4"
+
+S=${WORKDIR}/${PN}
+
+src_prepare() {
+	eqmake4 virtualjaguar.pro -o makefile-qt
```
mgorny: Wait a minute… why is this done in src_prepare()? It definitely belongs in src_configure()! That would explain why it was called twice :).

damex: yeah. it is.

PR #5, opened 2015-02-07
USE=libav on some packages

mgorny:
To make it easier for me to merge it into CVS :P. But feel free to review.


mgorny:
Merged via CVS :).

PR #7, opened 2015-02-08
fix typo in metasploit metadata.xml - bug #533764

mm1ke:
Hi! 

This is my first pull request and i hope it's OK to pull changes were a bug [1] is already open. Since i didn't got any response on my bug and the fix is rather trivial i though to go via github. 

I also would like to create more pull requests in the future, including changes which i already posted on bugs.gentoo.org. However, first, is it even a good idea to create patches on two places (b.g.o and github) and secondly is there a rule what should go to github and what to b.g.o?
Since gentoo provides a gentoo rsync mirror i would rather fix things via github instead of posting a patch on b.g.o.

[1] https://bugs.gentoo.org/show_bug.cgi?id=533764

### DIFF
--- a/net-analyzer/metasploit/metadata.xml
+++ b/net-analyzer/metasploit/metadata.xml
@@ -7,7 +7,7 @@
 		<description>Very delicate package, ask before you touch please</description>
 	</maintainer>
 	<use>
-		<flag name='development'>Install dependancies needed for metasploit and exploit development</flag>
+		<flag name='development'>Install dependencies needed for metasploit and exploit development</flag>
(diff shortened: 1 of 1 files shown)


mgorny:
Thanks, merged now.

And yes, it's fine to submit stuff both in PR and bugs. Bugs are somewhat more official and a good reference point for bigger changes. For small fixes (like this typo fix), bugs aren't really necessary. For bigger ones, they're good to have so people can yell at you somewhere officially :).

The point of the mirror is to let you submit fixes in a way that's convenient to you :). I also find it convenient to merge Pull Requests, compared to working with bugzilla and patches.

PR #8, opened 2015-02-08
net-libs/obby - depend on net/avahi[dbus] for libavahi-client (bug #473740), update header

mm1ke:
Pretty old bug [1] but still valid. (tested today)
Would be great if we could patch the ebuild in order to fix this bug.
I've also updated the header (year) in order to make repoman happy :)

[1] https://bugs.gentoo.org/show_bug.cgi?id=473740


mgorny:
Ok, I can confrim that `avahi[-dbus]` doesn't install libavahi-client. Funny design.

mgorny:
No need to update years btw, echangelog does that when we commit :).

mgorny:
Merged now.

mm1ke:
thx, good to know - won't update the year in the future :)

PR #9, opened 2015-02-08
games-fps/wordofpadman - depend on libsdl[X], update header

mm1ke:
Another old but still valid bug [1](again, tested today)
The fix is simple, just depend on media-libs/libsdl[X], also updated the header.

[1] https://bugs.gentoo.org/show_bug.cgi?id=459360


mgorny:
I'm assuming the bug has been fixed and it works for you now. Feel free to reopen if it doesn't.

PR #11, opened 2015-02-10
net-analyzer/speedtest-cli - add setuptools dependency (bug# 502302)

mm1ke:
This annoyed me for quite some time. Especially on minimal systems, were setuptools wasn't pulled by any other packages, i've hit this quite often. Would be nice if we could finally add this dependency.

Gentoo-Bug: https://bugs.gentoo.org/show_bug.cgi?id=502302
Btw, the bug is for Version 0.2.4, but it's still valid for 0.2.5 - just testet it + see duplicate.

Maybe this should also block: https://bugs.gentoo.org/show_bug.cgi?id=530994

### DIFF
--- a/net-analyzer/speedtest-cli/speedtest-cli-0.2.5.ebuild
+++ b/net-analyzer/speedtest-cli/speedtest-cli-0.2.5.ebuild
@@ -1,4 +1,4 @@
-# Copyright 1998-2015 Gentoo Foundation
+# Copyright 1999-2015 Gentoo Foundation
 # Distributed under the terms of the GNU General Public License v2
 # $Header: /var/cvsroot/gentoo-x86/net-analyzer/speedtest-cli/speedtest-cli-0.2.5.ebuild,v 1.3 2015/01/24 14:12:18 ago Exp $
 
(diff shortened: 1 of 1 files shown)


net-analyzer/speedtest-cli/speedtest-cli-0.2.5.ebuild
```diff
@@ -1,4 +1,4 @@
-# Copyright 1999-2015 Gentoo Foundation
+# Copyright 1998-2015 Gentoo Foundation
```
mgorny: Please don't move copyright back :).

mgorny:
Please fix the copyright year and then I'll merge it.

prometheanfire:
dunno if we have this set up

prometheanfire:
test this please

mgorny:
Merged.

Now a quick extra lesson: in trivial-yet-important fixes like this, you may consider adding them to older stable version too. I just did that after confirming it uses setuptools as way. While you generally should be extra cautious when dealing with stable ebuilds, adding extra `DEPEND` to unbreak them is rather fine :).

PR #12, opened 2015-02-10
sys-fs/cryptsetup - fix typo in dmcrypt.confd (bug #518592)

mm1ke:
This is a very simple fix for a typo in the dmcrypt configuration file. "mount" should be replaced by "target" since mount isn't described anywhere in the configuration file and also doesn't work while the same example would work with "target". 

Gentoo Bug: https://bugs.gentoo.org/show_bug.cgi?id=518592


mgorny:
I'm going to trust you with this one :). Merged.

PR #13, opened 2015-02-10
media-video/rtmpdump - fix multilib build (bug #501098)

mm1ke:
This fixes multilib build for rtmpdump.
The actual stable version in tree is broken so i've just added the missing parts (note: the live version already includes multilib dependencies). I've also changed multilib_build_binaries with multilib_is_native_abi since the first one is deprecated.

Regarding zlib: Is it really needed to have multilib dependencies here as well? multilib rtmpdump seems to build fine even without multilib zlib.

I've test-build rtmpdump without any flags, ssl and ssl+gnutls enabled. However i didn't test building it with polarssl support.

Gentoo Bug: https://bugs.gentoo.org/show_bug.cgi?id=501098

### DIFF
--- a/media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
+++ b/media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
@@ -17,9 +17,9 @@ KEYWORDS="amd64 ~arm hppa ~mips ppc ppc64 x86 ~amd64-fbsd ~x86-fbsd ~amd64-linux
 IUSE="gnutls polarssl ssl"
 
 DEPEND="ssl? (
-		gnutls? ( net-libs/gnutls )
-		polarssl? ( !gnutls? ( >=net-libs/polarssl-0.14.0 ) )
-		!gnutls? ( !polarssl? ( dev-libs/openssl ) )
+		gnutls? ( net-libs/gnutls[${MULTILIB_USEDEP}] )
+		polarssl? ( !gnutls? ( >=net-libs/polarssl-0.14.0[${MULTILIB_USEDEP}] ) )
+		!gnutls? ( !polarssl? ( dev-libs/openssl[${MULTILIB_USEDEP}] ) )
 	)
 	sys-libs/zlib"
 RDEPEND="${DEPEND}"
@@ -60,7 +60,7 @@ multilib_src_compile() {
 	fi
 	#fix multilib-script support. Bug #327449
 	sed -i "/^libdir/s:lib$:$(get_libdir)$:" librtmp/Makefile
-	if ! multilib_build_binaries; then
+	if ! multilib_is_native_abi; then
 		cd librtmp
 	fi
 	emake CC="$(tc-getCC)" LD="$(tc-getLD)" \
--- a/media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
+++ b/media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
@@ -61,7 +61,7 @@ multilib_src_compile() {
 	#fix multilib-script support. Bug #327449
 	sed -i "/^libdir/s:lib$:$(get_libdir)$:" librtmp/Makefile
 	if ! multilib_is_native_abi; then
-		cd librtmp
+		cd librtmp || die
 	fi
 	emake CC="$(tc-getCC)" LD="$(tc-getLD)" \
 		OPT="${CFLAGS}" XLDFLAGS="${LDFLAGS}" CRYPTO="${crypto}" SYS=posix
--- a/media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
+++ b/media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
@@ -17,11 +17,11 @@ KEYWORDS="amd64 ~arm hppa ~mips ppc ppc64 x86 ~amd64-fbsd ~x86-fbsd ~amd64-linux
 IUSE="gnutls polarssl ssl"
 
 DEPEND="ssl? (
-		gnutls? ( net-libs/gnutls[${MULTILIB_USEDEP}] )
-		polarssl? ( !gnutls? ( >=net-libs/polarssl-0.14.0[${MULTILIB_USEDEP}] ) )
-		!gnutls? ( !polarssl? ( dev-libs/openssl[${MULTILIB_USEDEP}] ) )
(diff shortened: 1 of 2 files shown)


media-video/rtmpdump/rtmpdump-9999.ebuild
```diff
@@ -54,7 +54,7 @@ multilib_src_compile() {
 	fi
 	#fix multilib-script support. Bug #327449
 	sed -i "/^libdir/s:lib$:$(get_libdir)$:" librtmp/Makefile
-	if ! multilib_build_binaries; then
+	if ! multilib_is_native_abi; then
 		cd librtmp
```
mgorny: Add missing `|| die` here, please :). Preferably fix issues such as this on sight.

media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
```diff
@@ -17,11 +17,11 @@ KEYWORDS="amd64 ~arm hppa ~mips ppc ppc64 x86 ~amd64-fbsd ~x86-fbsd ~amd64-linux
 IUSE="gnutls polarssl ssl"
 
 DEPEND="ssl? (
-		gnutls? ( net-libs/gnutls )
-		polarssl? ( !gnutls? ( >=net-libs/polarssl-0.14.0 ) )
-		!gnutls? ( !polarssl? ( dev-libs/openssl ) )
+		gnutls? ( net-libs/gnutls[${MULTILIB_USEDEP}] )
```
mgorny: When adding multilib deps, you need to also add `>=` for version that introduced multilib support or EAPI 5. This is normally done using [multilib-dep-fixor](https://bitbucket.org/mgorny/multilib-dep-fixor) but in this case you can just copy them over from -9999.

Afterwards, please do `repoman full` scan again to ensure that all the deps are stable as well.

mgorny:
As for zlib, check which libraries 32-bit librtmp links to. You can use either `readelf`, `scanelf` or `lddtree` for that (but not `ldd` because it reports dependencies recursively). In fact, Gentoo `less` automatically calls readelf for that when called on .so.

mm1ke:
Thanks for the hints!
Didn't know about multilib-dep-fixor and honestly - i didn't look further than just fixing the dependency issues in the first place :)
However, i've added the || die and copied the deps from the 9999 ebuild now

About zlib:
I've checked the libraries and it looks like zlib is only used when rtmpdump is built with a cryptography library. 
lddtree shows: 

> obelix ~ # lddtree /usr/lib64/librtmp.so.1 
> librtmp.so.1 => /usr/lib64/librtmp.so.1 (interpreter => none)
>    libc.so.6 => /lib64/libc.so.6
>        ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
> obelix ~ # lddtree /usr/lib32/librtmp.so.1 
> librtmp.so.1 => /usr/lib32/librtmp.so.1 (interpreter => none)
>    libc.so.6 => /lib32/libc.so.6
>        ld-linux.so.2 => /lib64/ld-linux.so.2

BTW: I'm using "objdump -x ...|grep NEEDED" mostly ;)

From the HP: "To build with all features enabled, you will also need a Cryptography library and the zlib compression library." - guess that means, without any cryptography enabled - zlib won't be used too.

I've moved the dependency into ssl?( .... ) so it only gets pulled if a cryptography lib is enabled. 

What do you think about that?

mgorny:
Makefile agrees with you:

```
LIBZ=-lz
LIB_GNUTLS=-lgnutls -lhogweed -lnettle -lgmp $(LIBZ)
LIB_OPENSSL=-lssl -lcrypto $(LIBZ)
LIB_POLARSSL=-lpolarssl $(LIBZ)
```

Going further, you can look into replacing those LIB_\* thingies with proper pkg-config :). But maybe not in this Pull Request but when you're bored one day.

Now I'll review it.

media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
```diff
@@ -17,11 +17,11 @@ KEYWORDS="amd64 ~arm hppa ~mips ppc ppc64 x86 ~amd64-fbsd ~x86-fbsd ~amd64-linux
...
-	)
-	sys-libs/zlib"
+		gnutls? ( >=net-libs/gnutls-2.12.23-r6[${MULTILIB_USEDEP}] )
+		polarssl? ( !gnutls? ( >=net-libs/polarssl-1.3.4[${MULTILIB_USEDEP}] ) )
+		!gnutls? ( !polarssl? ( >=dev-libs/openssl-1.0.1h-r2[${MULTILIB_USEDEP}] ) )
+		>=sys-libs/zlib-1.2.8-r1[${MULTILIB_USEDEP}]
+	)"
 RDEPEND="${DEPEND}"
 
 pkg_setup() {
```
mgorny, line 27: In the line below you can replace subshell with:

```
if ! use ssl && { use gnutls || use polarssl; }; then
```

media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
```diff
@@ -60,8 +60,8 @@ multilib_src_compile() {
 	fi
 	#fix multilib-script support. Bug #327449
 	sed -i "/^libdir/s:lib$:$(get_libdir)$:" librtmp/Makefile
```
mgorny: Here `|| die` too.

media-video/rtmpdump/rtmpdump-2.4_p20131018.ebuild
```diff
@@ -60,8 +60,8 @@ multilib_src_compile() {
...
-	if ! multilib_build_binaries; then
-		cd librtmp
+	if ! multilib_is_native_abi; then
+		cd librtmp || die
```
mgorny, line 64: There are more `cd`s you didn't fix :).

mm1ke:
Embarrassing - i should really look closer at ebuilds...
Another round of "die" fixes + replaced subshell.

mgorny:
Well, great work, thanks ;). However, you didn't run repoman :P! If you did, you'd notice the failures in dependency tree. Those failures are likely the reason someone didn't add MULTILIB_USEDEPs there in the first place, though I prefer if people ask or file bugs rather than sweeping issues under the carpet. But I'm digressing now.

I've fixed the profiles to make repoman happy and merged your commits.

mm1ke:
Thanks! - and sorry for not running repoman. Forget that.. :)

PR #14, opened 2015-02-11
Use pcheck instead of repoman

mgorny:
Unfinished and I'm going to sleep. I'll fix it tomorrow unless someone beats me to it.

@mrueg, @junghans.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: python
 python:
-  - pypy
+  - 2.7
+
 env:
+  - JOB=global
   - JOB=0
   - JOB=1
   - JOB=2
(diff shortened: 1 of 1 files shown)


junghans:
```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).
```

mgorny:
It's getting better and better now. Now I'm working on fixing the checks to be correct.

mgorny:
We need to wait for @radhermit to fix pkgcore wrt implicit IUSE.

mgorny:
I guess there's no point waiting more. Better have a 'mostly' working CI that will error out at few last packages than none at all.

@swegener, could you merge this, please?

PR #16, opened 2015-02-13
net-misc/electrum - add media-gfx/zbar[v4l] dependency (bug #539984)

mm1ke:
This fixes gentoo bug 539984[1].
I've also added the dependency to version 1.9.7 as this version is also affected. Older versions of electrum doesn't have a qrcode use flag.

repoman seems to be happy, just complains about the invalid copyright - which should get updated anyway...

[1]https://bugs.gentoo.org/show_bug.cgi?id=539984


mgorny:
```
+  14 Feb 2015; Michał Górny <mgorny@gentoo.org> electrum-1.9.7.ebuild,
+  electrum-1.9.8.ebuild:
+  USE=qrcode requires zbar[v4l], bug #539984 with PR by Michael Mair-Keimberger.
```

Thanks a lot for all your work!

PR #18, opened 2015-02-15
sys-power/ncpufreqd - fix homepage (bug #516802)

mm1ke:
Another simple homepage fix - see bug[1]

Note:
The original download link is also dead, so i've changed that to the bitbucket link. However, the tarball size changed with that. I've compared both tarballs (the old from our mirrors) and it's basically the same tarball. The difference is that the old one has ncpufreqd compiled in the tarball (32bit) and contains a Makefile (but since they changed to cmake it's probably not needed anymore..)
the newer one also contain a ".hgtags" file.

After all, this means our mirror sources needs to be updated as well.

[1]https://bugs.gentoo.org/show_bug.cgi?id=516802


sys-power/ncpufreqd/ncpufreqd-2.4.ebuild
```diff
@@ -7,8 +7,8 @@ EAPI="5"
 inherit cmake-utils
 
 DESCRIPTION="Daemon controlling CPU speed and temperature"
-HOMEPAGE="http://projects.simpledesigns.com.pl/project/ncpufreqd/"
-SRC_URI="http://projects.simpledesigns.com.pl/get/${P}.tar.bz2"
+HOMEPAGE="https://bitbucket.org/nelchael/ncpufreqd"
+SRC_URI="https://bitbucket.org/nelchael/${PN}/get/${P}.tar.bz2"
```
mgorny: Now here's the tricky part: we can't really use tag-generated tarballs like this. They may be generated on the fly, and so you can't trust having the same checksum on every download.

Now, if you were a developer I'd just ask you to mirror the file in your devspace. However, you can just ask Krzysiek to provide a downloadable tarball on bitbucket or elsewhere. He used to be a Gentoo developer so he should understand the issue :).

mm1ke: I've sent an email. I'll wait and see how he respond :)

mgorny: No reply yet?

mm1ke: Unfortunately, no.

mgorny: Tell you what.

```
SRC_URI="http://dev.gentoo.org/~mgorny/dist/${P}.tar.xz"
```

Please let me know if I did everything right. And remember to update DEPEND for .tar.xz.

mgorny:
We are required to explicitly depend on any unpacker other than gzip & bzip2. That's how PMS states it.

mgorny:
Sorry, i'm wrong here. PMS says we need to ensure it somehow. I think we had a policy that required explicit xz-utils but I'm not sure if that's still valid.

mgorny:
Merged.

kensington:
I believe the policy regarding xz-utils was for before it was in the system profile. Some people prefer to keep the explicit depend anyway and that's pefectly valid too.

PR #21, opened 2015-02-16
media-gfx/nvidia-cg-toolkit - improve ebuild

mm1ke:
This PR tries fixing two bugs:
First - it updates the HP in both ebuilds [1]
Secondly, i've tried to improve/fix it's dependencies. [2]

Since the ebuild already got updated for multilib, i haven't look closer on the rest of the ebuild. However, the original ebuild included two dependencies which i couldn't find were they should be used. For now i've decided to keep them. I would like to hear other opinions about them.

media-libs/freeglut:
I haven't found any lib/binary which directly depends on freeglut. (i've also checked all its examples). However, /local/Cg/examples/README lists freeglut as dependency (even though it seems like they shipping freeglut with the tarball anyway  - with its examples, see /home/michael/nvidia-cg/local/Cg/examples/Trace/)

virtual/libstdc++:3.3:
Don't know about this. This might be correct, but i don't know how to check this and since i haven't set up a x86 system atm - someone else might be able to clear this up :)

The rest of the deps are mainly coming from the examples, but also /bin/cgfxcat depends on all those x11-libs.
Please review :)

btw, repoman complains about the emul deps:

>  dependency.unknown            1
>   media-gfx/nvidia-cg-toolkit/nvidia-cg-toolkit-3.1.0013-r3.ebuild: RDEPEND: app-emulation/emul-linux-x86-opengl[-abi_x86_32(-)], app-emulation/emul-linux-x86-xlibs[-abi_x86_32(-)]

Don't know exactly how to fix this... :)

[1] https://bugs.gentoo.org/show_bug.cgi?id=518144
[2] https://bugs.gentoo.org/show_bug.cgi?id=466832


mgorny:
Ok, this one's going to require more maintainer attention.

@lu-zero, @maekke, @radhermit, @aballier, any of you would like to review this? Though I feel that a package like this would use a more specific maintainer than graphics herd. @mm1ke, would you like to proxy-maintain it? :)

mm1ke:
Sure, i guess i could do that :)

jlec:
Thanks for your contribution!

```
+*nvidia-cg-toolkit-3.1.0013-r4 (19 Feb 2015)
+
+  19 Feb 2015; Justin Lecher <jlec@gentoo.org>
+  +nvidia-cg-toolkit-3.1.0013-r4.ebuild, metadata.xml,
+  nvidia-cg-toolkit-3.1.0013-r2.ebuild, nvidia-cg-toolkit-3.1.0013-r3.ebuild:
+  Merge PR, thanks mm1ke for preperation, fixes #518144 & #466832; correct
+  QA_PREBUILT for USE=examples; make exmaples executable; drop unnecessary deps
+
```

PR #22, opened 2015-02-17
boot0,freebsd-{lib,libexec}: fix bug 511698, it's a critical bug.

nigoro:
Hello.
This is a request to fix the critical issue that reported in May last year.
If patch set is not applied, it is possible that all of the command is unexecutable.

boot0,sys-freebsd/freebsd-{lib,libexec}: compile with gcc-4.8.3 or later, ALL COMMANDS RETURNS Abort trap: 6 (core dumped)
https://bugs.gentoo.org/show_bug.cgi?id=511698

### DIFF
--- a/sys-freebsd/boot0/files/boot0-add-nossp-cflags.patch
+++ b/sys-freebsd/boot0/files/boot0-add-nossp-cflags.patch
@@ -1,0 +1,11 @@
+https://bugs.gentoo.org/show_bug.cgi?id=511698
+
+diff --git a/sys/boot/Makefile.inc b/sys/boot/Makefile.inc
+index e0039b9..533dea0 100644
+--- a/sys/boot/Makefile.inc
++++ b/sys/boot/Makefile.inc
+@@ -1,3 +1,3 @@
+ # $FreeBSD: release/10.0.0/sys/boot/Makefile.inc 188895 2009-02-21 15:04:31Z ru $
+ 
+-SSP_CFLAGS=
++SSP_CFLAGS= -fno-stack-protector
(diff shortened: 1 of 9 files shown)


sys-freebsd/boot0/files/boot0-add-nossp-cflags.patch
```diff
@@ -0,0 +1,11 @@
+https://bugs.gentoo.org/show_bug.cgi?id=511698
```
mgorny, line 1: I'll merge it as-is because of the urgency but next time (or maybe fix it later), please put some textual explanation. Pure links aren't good since they require accessing external resource.

sys-freebsd/freebsd-lib/files/freebsd-lib-add-nossp-cflags.patch
```diff
@@ -0,0 +1,30 @@
+Please do not disable this patch.
```
mgorny, line 1: Oh, here's an explanation but not a good one. Focus on what the patch changes, no need for big fat warnings when patch is described correctly.

mgorny:
```
+  18 Feb 2015; Michał Górny <mgorny@gentoo.org>
+  +files/boot0-add-nossp-cflags.patch, boot0-9.1.ebuild, boot0-9.2.ebuild:
+  Fix SIGABRT failure when compiled with SSP-enabled gcc, bug #511698, patch by
+  Yuta SATOH

+  18 Feb 2015; Michał Górny <mgorny@gentoo.org>
+  +files/freebsd-lib-add-nossp-cflags.patch, freebsd-lib-9.1-r11.ebuild,
+  freebsd-lib-9.2.ebuild:
+  Fix SIGABRT failure when compiled with SSP-enabled gcc, bug #511698, patch by
+  Yuta SATOH

+  18 Feb 2015; Michał Górny <mgorny@gentoo.org>
+  +files/freebsd-mk-defs-add-nossp-cflags.patch, freebsd-mk-defs-9.1.ebuild,
+  freebsd-mk-defs-9.2.ebuild:
+  Fix SIGABRT failure when compiled with SSP-enabled gcc, bug #511698, patch by
+  Yuta SATOH
```

PR #25, opened 2015-02-19
#FL-2108 && upstream #89213: make pulseaudio's udev use require alsa or oss use

damex:
making pulseaudio's udev use depend on/require alsa or oss use as it is described/declared at the sources.

credits for resolving this bug goes to @mgorny and Nexie Kind from funtoo-users.


mgorny:
Thanks a lot, merged now.

```
+  19 Feb 2015; Michał Górny <mgorny@gentoo.org> pulseaudio-5.0-r4.ebuild,
+  pulseaudio-5.0-r5.ebuild, pulseaudio-5.0-r6.ebuild, pulseaudio-5.0-r7.ebuild,
+  pulseaudio-5.99.3.ebuild:
+  Add REQUIRED_USE for USE=udev since it requires either alsa or oss being
+  enabled. https://bugs.funtoo.org/browse/FL-2108 with investigation by Nexie
+  Kind and pull request by Roman Kuzmitsky.
```

PR #26, opened 2015-02-19
app-portage/portage-utils - fix homepage (bug #534468)

mm1ke:
Another simple HP fix :)

The ebuild(s) however could get some more love, (EAPI bump for example) but i've decided to start with the HP since other changes probably imply an revision bump. (i could look into into it for another PR)


mgorny:
Not sure about this. The wiki page isn't really the homepage, it's some random doc describing those tools. I'm going to ask more…

mm1ke:
Hmm, another possible homepage could be portage-utils git repository which i think would be [1]. However, since the actual ebuild already points to a wiki page, i though it was ok and so i've just updated the link. (the actual link just points to an old wiki page - which redirects to the current one)

BTW, i hope it's ok to change the hp for all ebuilds? I couldn't find any information what todo if a homepage moves (like updating only the latest or all ebuilds...)

[1] http://cgit.gentooexperimental.org/proj/portage-utils.git/

mgorny:
Merged.

PR #27, opened 2015-02-20
qca gpg plugin fake use flag

damex:


kensington:
@Pesa maybe you have some comments? The idea of the fake USE flags is to avoid blockers difficult for portage to resolve.

app-crypt/qca/qca-2.0.3-r1.ebuild
```diff
@@ -0,0 +1,80 @@
...
+
+inherit eutils multilib qt4-r2
+
+DESCRIPTION="Qt Cryptographic Architecture (QCA)"
+HOMEPAGE="http://delta.affinix.com/qca/"
+SRC_URI="http://delta.affinix.com/download/${PN}/${PV%.*}/${P}.tar.bz2"
+
+LICENSE="LGPL-2"
+SLOT="2"
+KEYWORDS="alpha amd64 ~arm hppa ~ia64 ppc ppc64 sparc x86 ~amd64-fbsd ~x86-fbsd ~amd64-linux ~x86-linux ~ppc-macos ~sparc-solaris"
```
mgorny: You have to drop stable keywords to unstable when bumping. Repoman doesn't allow you to commit straight to stable. This can be done e.g. using:

```
ekeyword ~all qca-2.0.3-r1.ebuild
```

johu:
EAPI3 is deprecated. Please move to newest.

mgorny:
@pesa, to expand on this more.

```
|| ( qca[gpg] qca-gnupg )
```

causes Portage to consider `qca-gnupg` even when new enough `qca[-gpg]` is installed which results in awful blockers. This, combined with plain `qca[gpg]` deps would mean that every version of qca is installed fine without blockers. Users of old versions will be asked to set the flag early but they would eventually have to anyway.

mgorny:
@johu, this is going almost straight to stable. Changing EAPI is not the best idea, and a waste of time consider the ebuild's eventually going to be removed.

johu:
@mgorny why not stabilize newer version instead?

Pesa:
I don't have any particular comment... TBH I don't really care about qca since I stopped using `net-im/psi` years ago.

Just a clarification:

> Users of old versions will be asked to set the flag early but they would eventually have to anyway.

They will be asked because of what? qca consumers?

mgorny:
@johu, could be done. Though it would expose stable users to the issue for some time. Doing this on old version has the advantage that both branches of `|| ()` would be fine and blocker-free.

@pesa, yes, if something depended on `qca[gpg]` or like, Portage will ask to add `qca gpg` to package.use.

mgorny:
```
+*qca-2.0.3-r1 (21 Feb 2015)
+
+  21 Feb 2015; Michał Górny <mgorny@gentoo.org> +qca-2.0.3-r1.ebuild:
+  Add USE flags for plugins that were integrated in the newer qca version. The
+  USE flags pull the plugins implicitly making it easier to create useful deps
+  on qca with plugins. Patch prepared by Roman Kuzmitsky (damex).
```

Along with some fancy arch profile masks for non-keyworded plugins.

PR #28, opened 2015-02-21
Perl revbump

mgorny:
Because @dilfridge failed to bump the virtuals when changing deps and this gives a lot of slot conflicts.


mgorny:
And committed.

PR #29, opened 2015-02-21
app-benchmark/stress - fix homepage (bug #538356), raise EAPI, fix download link, fix LICENCE

mm1ke:
Another HP fix with some additional changes:

As in bug 538356 [1] it seems like the HP changed which i've fixed in all ebuild's. However, at the new HP i could only find sources for the latest version which also differs from our sources. While it seems like the sourcecode didn't changed at all it seems like it got re-licensed. (see COPYING)
Because of that i've changed the Licence for 1.0.4 to GPL-3 which might legitimate a revision bump.

[1] https://bugs.gentoo.org/show_bug.cgi?id=538356


mgorny:
So old versions are unfetchable now? We may want for stabilization to finish and remove them.

mm1ke:
Well, they are still on our mirrors. I could replace SRC_URI with mirror://gentoo/${P}.tar.gz

mgorny:
Nah, no need to. However…

app-benchmarks/stress/stress-1.0.4.ebuild
```diff
@@ -2,16 +2,16 @@
...
+EAPI=5
 
 inherit flag-o-matic
 
 MY_P=${PN}-${PV/_/}
 DESCRIPTION="Imposes stressful loads on different aspects of the system"
-HOMEPAGE="http://weather.ou.edu/~apw/projects/stress"
-SRC_URI="http://weather.ou.edu/~apw/projects/stress/${MY_P}.tar.gz"
+HOMEPAGE="http://people.seas.harvard.edu/~apw/stress"
+SRC_URI="http://people.seas.harvard.edu/~apw/stress/${MY_P}.tar.gz"
```
mgorny: Since the file contents have changed (due to re-license), you have to revbump the ebuild and rename the file, e.g.:

```
SRC_URI="http://people.seas.harvard.edu/~apw/stress/${MY_P}.tar.gz -> ${MY_P}-r1.tar.gz"
```

So that the new version can replace the old one cleanly.

Also, is this just regular author-move or some kind of fork? Because if the latter, I'm not sure if the license change is fully legal.

mm1ke:
Ok, i've added an r1 of stress.
About the licence i'm not quite sure. However, Amos Waterland, the owner of the HP, is also the maintainer of stress. I guess it's rather a regular author-move than a fork..

mgorny:
Thanks a lot.

```
+*stress-1.0.4-r1 (22 Feb 2015)
+
+  22 Feb 2015; Michał Górny <mgorny@gentoo.org> +stress-1.0.4-r1.ebuild,
+  stress-0.18.6.ebuild, stress-1.0.0.ebuild, stress-1.0.1.ebuild,
+  stress-1.0.3.ebuild, stress-1.0.4.ebuild:
+  Fix HOMEPAGE, bug #538356. Revbump for 1.0.4 was re-released with GPL-3
+  license. https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/29 by
+  mm1ke.
```

PR #31, opened 2015-02-21
sys-process/pidof-bsd-20050501-r4: fix compile fails on FreeBSD 10.0

nigoro:
NOMAN is not longer supported. Need to change to NO_MAN.

https://bugs.gentoo.org/show_bug.cgi?id=483044

NOTE,
NO_MAN is already supported on FreeBSD 8.0.


mgorny:
```
+  21 Feb 2015; Michał Górny <mgorny@gentoo.org>
+  +files/pidof-bsd-20050501-fbsd10.patch, pidof-bsd-20050501-r4.ebuild:
+  Fix build failure on FreeBSD 10.0, https://github.com/gentoo/gentoo-portage-
+  rsync-mirror/pull/31 by nigoro.
```

PR #32, opened 2015-02-21
[app-crypt/qca] transfer stable keywords

damex:


mgorny:
Oh, I merged this one already.

PR #33, opened 2015-02-22
app-arch/star - update homepage/SRC_URI (bug #537816)

mm1ke:
Simple fix for bug 537816 [1]. Fetches, compiles and repoman is happy too :)

[1]https://bugs.gentoo.org/show_bug.cgi?id=537816


jlec:
-  22 Feb 2015; Justin Lecher jlec@gentoo.org star-1.5.2.ebuild:
-  Move ebuild from berlios.de to sf.net, #537816; thanks mm1ke for the patch
  +

PR #34, opened 2015-02-22
net-irc/ngircd - remove berlios link from SRC_URI (bug #537818)

mm1ke:
Another SRC_URI fix (bug 537818). I've only removed the berlios link and raised EAPI for the last 3 versions.

[1] https://bugs.gentoo.org/show_bug.cgi?id=537818


net-irc/ngircd/ngircd-17.1.ebuild
```diff
@@ -2,14 +2,13 @@
 # Distributed under the terms of the GNU General Public License v2
 # $Header: /var/cvsroot/gentoo-x86/net-irc/ngircd/ngircd-17.1.ebuild,v 1.5 2014/03/10 10:58:31 ssuominen Exp $
 
-EAPI=4
+EAPI="5"
```
mgorny, line 5: You should rather avoid raising EAPI in stable ebuilds. Just in case the change could trigger some conditional change in eclass, and break stuff.

mgorny, line 5: Though here it looks fine, so I'll just merge it :P.

mgorny:
```
+  22 Feb 2015; Michał Górny <mgorny@gentoo.org> ngircd-0.12.1.ebuild,
+  ngircd-13.ebuild, ngircd-17.1.ebuild, ngircd-18.ebuild, ngircd-20.3.ebuild:
+  Remove berlios SRC_URI, bump EAPI. https://github.com/gentoo/gentoo-portage-
+  rsync-mirror/pull/34 by Michael Mair-Keimberger.
```

PR #36, opened 2015-02-22
freebsd.eclass: Fix for FreeBSD10.0 or later

nigoro:
-9999 ebuild, changed to use to force /usr/share/mk.
Fixed a problem of numerical comparison in 10.0.

Details please see
https://bugs.gentoo.org/show_bug.cgi?id=488214#c23
https://bugs.gentoo.org/show_bug.cgi?id=488214#c24

### DIFF
--- a/eclass/freebsd.eclass
+++ b/eclass/freebsd.eclass
@@ -129,6 +129,12 @@ freebsd_rename_libraries() {
 freebsd_src_unpack() {
 	if [[ ${MY_PV} == *9999* ]]; then
 		S="${WORKDIR}" subversion_src_unpack
+
+		# When share/mk exists in ${WORKDIR}, it is used on FreeBSD 10.0
+		# Removed "${WORKDIR}"/share/mk/*.mk, use to force /usr/share/mk.
+		if [[ ${PN} != freebsd-mk-defs ]] ; then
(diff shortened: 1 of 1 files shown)


eclass/freebsd.eclass
```diff
@@ -141,7 +147,8 @@ freebsd_src_unpack() {
 
 	# Starting from FreeBSD 9.2, its install command supports the -l option and
 	# they now use it. Emulate it if we are on a system that does not have it.
-	if [[ ${RV} > 9.1 ]] && ! has_version '>=sys-freebsd/freebsd-ubin-9.2_beta1' ; then
+	version_compare ${RV} 9.1
```
mgorny: Please avoid `version_compare` since it's hard to read. Prefer `! version_is_at_least`.

nigoro:
@mgorny, Thank you for advice. I have updated the patch.

The results of the test are as follows. 
9.1 --> False
9.2 --> True
10.1 --> True

mgorny:
```
+  22 Feb 2015; Michał Górny <mgorny@gentoo.org> freebsd.eclass:
+  Fix support for FreeBSD 10.0. Force /usr/share/mk there, and fix version
+  comparison for install commands.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/36 by nigoro.
```

PR #45, opened 2015-02-23
clang profiles: fix can't create stage1, and cleanup

nigoro:
Details, please see https://bugs.gentoo.org/show_bug.cgi?id=539858

### DIFF
--- a/profiles/arch/amd64-fbsd/clang/make.defaults
+++ b/profiles/arch/amd64-fbsd/clang/make.defaults
@@ -6,3 +6,6 @@ CFLAGS="-O2 -pipe"
 CXXFLAGS="-stdlib=libc++ ${CFLAGS}"
 FFLAGS="${CFLAGS}"
 FCFLAGS="${CFLAGS}"
+
+# sys-devel/clang require static-analyzer, bug 539858
+BOOTSTRAP_USE="${BOOTSTRAP_USE} static-analyzer"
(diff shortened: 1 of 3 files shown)


profiles/arch/amd64-fbsd/clang/make.defaults
```diff
@@ -6,3 +6,6 @@ CFLAGS="-O2 -pipe"
 CXXFLAGS="-stdlib=libc++ ${CFLAGS}"
 FFLAGS="${CFLAGS}"
 FCFLAGS="${CFLAGS}"
+
+# sys-devel/clang require static-analyzer, bug 539858
+BOOTSTRAP_USE="${BOOTSTRAP_USE} static-analyzer"
```
mgorny: Why `BOOTSTRAP_USE`? Wouldn't it be better to just `package.use.force` it on LLVM, like you do for clang?

nigoro: I used it because it was a problem of stage1.

sys-devel/llvm is enabled static-analyzer by default.
However, it is disabled by USE="-*" when making stage1.

Please comment in bugzilla. Might get a more detailed comments by aballier.

mgorny: Commented.

mgorny: Ping? I'm still waiting for reply there :P.

mgorny:
Merged, thanks.

PR #50, opened 2015-02-23
x11-themes/gtk-engines-candido - fix HP and SRC_URI (bug #537812)

mm1ke:
Next one [1]. I hope you don't mind the quotation marks for MY_P. It just feels more correct ;)
Also a eapi bump could possibly be done, but it's a stable ebuild and didn't wanted to create a -r1 just for EAPI=5.

[1]https://bugs.gentoo.org/show_bug.cgi?id=537812


mgorny:
```
+  23 Feb 2015; Michał Górny <mgorny@gentoo.org>
+  gtk-engines-candido-0.9.1.ebuild:
+  Update SRC_URI & HOMEPAGE. Bug #537812. https://github.com/gentoo/gentoo-
+  portage-rsync-mirror/pull/50 by mm1ke.
```

PR #51, opened 2015-02-23
Bug #537814 x11 themes slim themes

mm1ke:
More SRC_URI/HOMEPAGE fixes [1].

Note: While the new sourceforge links working fine i've encounter another small issue: The ebuild also fetches 3 files from www.konstantinhansen.de were the filesize are different from those of our mirrors. However, it doesn't seems like they changed. The only difference was that those from our mirror were packaged _with_ the directory included, while the files from konstantinhansen weren't.

Either we update files on our mirrors, or repack them somewhere else and update the links. What would be the best solution?

[1]https://bugs.gentoo.org/show_bug.cgi?id=537814


mgorny:
I'd say -r1 the three files if you're sure they are the same.

mgorny:
it was merged.

PR #53, opened 2015-02-23
sys-apps/gscanbus - fix HP and SRC_URI (bug # 537798)

mm1ke:
Another very simple SRC_URI/HOMEPAGE fix [1]. EAPI bump could be done as well(?) but i decided to leave it as is since it's a stable ebuild.

[1] https://bugs.gentoo.org/show_bug.cgi?id=537798


mgorny:
```
+  23 Feb 2015; Michał Górny <mgorny@gentoo.org> gscanbus-0.8.ebuild:
+  Fix HOMEPAGE & SRC_URI, bug #537798. https://github.com/gentoo/gentoo-portage-
+  rsync-mirror/pull/53 by mm1ke.
```

PR #55, opened 2015-02-24
net-p2p/gift-fasttrack - fix SRC_URI (bug #537784)

mm1ke:
Next patch [1]. Also added quotes to make repoman happy.

[1] https://bugs.gentoo.org/show_bug.cgi?id=537784


mgorny:

PR #56, opened 2015-02-24
net-p2p/gift-ares - fix HOMEPAGE and SRC_URI (bug #537782)

mm1ke:
Pretty much the same like the previous PR [1]. i've removed COPYING from dodoc as repoman was not happy about it.

[1]https://bugs.gentoo.org/show_bug.cgi?id=537782


mgorny:

PR #57, opened 2015-02-24
media-sound/horgand - fix HOMEPAGE and SRC_URI (bug #537762)

mm1ke:
Simple HOMEPAGE / SRC_URI fix [1]

[1] https://bugs.gentoo.org/show_bug.cgi?id=537762


mgorny:

PR #61, opened 2015-02-26
[net-im/telepathy-connection-managers]: add gadu, meanwhile and steam USE flags

karolherbst:
This will makes it possible for ktp-account-kcm in the kde overlay to depend on those services a clean way. Also this should make other users happy who want to use these services through telepathy.

Package-Manager: portage-2.2.17


profiles/arch/amd64/package.use.mask
```diff
@@ -18,6 +18,10 @@
 
 #--- END OF EXAMPLES ---
 
+# Karol Herbst <gentoo@karolherbst.de>
```
mgorny: No date, no love ;P. But don't hurry yet, I need to discuss with others if we can allow non-gentoo addresses here :).

karolherbst: no problems.

mgorny: Ok, we figured out that my name will be in ChangeLogs, so it should be here as well. So please change that to:

```
# Michał Górny <mgorny@gentoo.org> (01 Mar 2015)
```

and I shall be able to proceed :).

karolherbst: done

mgorny:
Merged, thanks.

PR #63, opened 2015-03-01
xfce4 4.12 bump

damex:
we're bumping xfce4 to 4.12 version.


mgorny:
Done. Had to add missing use descriptions to metadata.xml and mask xfce4-settings[libinput] on a few arches.

PR #64, opened 2015-03-01
add multilib and gtk2/gtk3 multibuild libappindicator

karolherbst:
fixes https://bugs.gentoo.org/show_bug.cgi?id=462764

also enables the possibilities to have systray icons for gtk2 apps in plasma5


karolherbst:
forgot repoman

PR #65, opened 2015-03-01
add multilib and gtk2/gtk3 multibuild libappindicator

karolherbst:
fixes https://bugs.gentoo.org/show_bug.cgi?id=462764

also enables the possibilities to have systray icons for gtk2 apps in plasma5

TODOs:
- [ ] "Conditional patching? No. Put a proper conditional there, and pass it to configure. Make it upstreamable!"
- [x] fix submenu for steam


mgorny:
@kensington, @alexxy, @creffett, @ABCDwp, @jmbsvicetto, @johu, @mrueg, @scarabeusiv,

mrueg:
I can't comment on the correctness of the multilib changes, but LGTM.
This is necessary to support systray icons for gtk+:2 based packages in plasma 5, thus please merge it after review.

johu:
EAPI 5 please

karolherbst:
@johu done

kensington:
Seems all good from KDE team side.

karolherbst:
I think @anyc will be interessted in this too, because inside the steam-overlay are some ebuilds for gtk2 supporti installing inside /opt

karolherbst:
okay, I've found an issue here: usex prints no if a USE flag isn't enabled.

karolherbst:
fixed

karolherbst:
I fixed most of the stuff @mgorny mentioned. I left a TODO in the PR description, which should be fixed as well.

karolherbst:
for reference:
@mgorny left some comments here:
https://github.com/karolherbst/gentoo-portage-rsync-mirror/commit/4ba0bdf2e4fc3e871a53f105f5745f7cff92dd32
https://github.com/karolherbst/gentoo-portage-rsync-mirror/commit/c4cffe636543c66ae5a5757fef03c8298a15f735
https://github.com/karolherbst/gentoo-portage-rsync-mirror/commit/d2f9f523dc9c5bb3bcc893c929b07373348f3791
https://github.com/karolherbst/gentoo-portage-rsync-mirror/commit/210cace4a0628d15db22d1c44dc7fbee4e0eb916
https://github.com/karolherbst/gentoo-portage-rsync-mirror/commit/de955fea35b9bc7c5d357beebb1f964cac406357

yngwin:
I've said this before on IRC: I really dislike the revision number abuse. I know this is the way the gnome team wants to do things, but besides being ugly and unnecessary, in my opinion it is wrong. If this were a package I would be co-maintaining, then I would veto this change.
Can we at least _try_ to get this past the ayatana team as a unified ebuild, with useflags to select gtk2/3, as is the proper way?

karolherbst:
I think the problem especially here is, that the packages don't allow to build gtk2 and gtk3 stuff in the same build, which means we would need multibuild. If we also add multilib stuff gets more complicated, because both have to be nested :/

karolherbst:
@mgorny how "severe" is the patch issue for you? I hardly have any experience with automake and don't think I will be able to create any "upstreamable" patches in the near future for it , so if somebody wants to pick this issue up, that's fine.

mgorny:
Not severe but I believe you can learn that :P.

mrueg:
Well I guess at least it is not merge-critical. Is there anything else that should be considered before merge?

karolherbst:
There is an issue with the steam systray menu anyway and I have to investigate it.

karolherbst:
so it works now.

mgorny:
Want me to review it? :P

mgorny:
2 merged, 5 left. I'd honestly wait a bit before adding GTK+2 so that people have time to change their USE=gtk deps before we change the meaning.

karolherbst:
okay, then I will rebase this on master

mgorny:
Thanks. If you have some time, could you try to figure out which overlays are still relying on `[gtk]` meaning gtk3? I guess zugaina can help you a bit.

mgorny:
Ok, sorry for waking up this late but doesn't this particular patch cause libdbusmenu-glib and some other stuff to be built twice? If it does, then please use some VPATH and other hackery to use 'master' build without gtk, and two 'slave' builds with gtk2/gtk3. Look at `net-libs/gtk-vnc` for a creepy example.

mgorny:
As a temporary solution, we may instead just force `?? ( gtk gtk3 )` since there's not many packages requiring gtk3 libdbusmenu.

karolherbst:
yeah, I know. this is really painful somehow :/ Even ubuntu is building it twice

mgorny:
It's minor concern for binary distros since developers are doings the builds there. In Gentoo, users are doing the builds.

karolherbst:
I think the best solution (also for upstream) would be to enable a build to compile against gtk2 and gtk3 together or let the gtk build use a system installed glib one.

mgorny:
```
+*libdbusmenu-12.10.2-r2 (19 Apr 2015)
+
+  19 Apr 2015; Michał Górny <mgorny@gentoo.org>
+  +files/libdbusmenu-12.10.2-gtk2-signal-fix.patch,
+  +libdbusmenu-12.10.2-r2.ebuild, metadata.xml:
+  Enable optional GTK+2 support. Enable multilib support. Remove gtk-doc install
+  hacks, just install into the proper gtk-doc location. Based on the work done
+  by Karol Herbst on https://github.com/gentoo/gentoo-portage-rsync-
+  mirror/pull/65. Part of bug #462764.
```

More hacky than your version but builds the glib part only once.

karolherbst:
so both libdbusmenu commits are merged now?

mgorny:
Yes but please test and confirm that my ebuild works well. This leaves us with the easier-but-ugly ebuilds.

karolherbst:
It seems to work here

mgorny:
Would you like to rebase again? We'd do a second review of the remaining stuff, and complain about it a lot :)

karolherbst:
guess what, I just forgot to push

mgorny:
Isn't pushing what makes people happy? :D

mgorny:
Well, libindicator looks rather clean, libappindicator will need more work to figure out how to make it saner. Supposedly some configure overrides instead of patching.

karolherbst:
yeah, libindicator was nice to handle, libappindicator ist a mess though, because of this python patch :/

karolherbst:
maybe it would make sense to use snapshot versions like used in the current LTS ubuntu or 15.04/14.10?

mrueg:
So can we get the outstanding commits into the tree? Even if the python patch is not the best way to solve it..

mgorny:
```
+*libindicator-12.10.1-r301 (07 May 2015)
+*libindicator-12.10.1-r201 (07 May 2015)
+
+  07 May 2015; Michał Górny <mgorny@gentoo.org>
+  +libindicator-12.10.1-r201.ebuild, +libindicator-12.10.1-r301.ebuild:
+  Add multilib support, thanks to Karol Herbst. Part of
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/65.
```

mgorny:
All done now, thanks. And please test :).

PR #66, opened 2015-03-01
net-wireless/python-wifi - update SRC_URI (bug #537788), raise eapi, use distutils-r1

mm1ke:
This PR fixes bug#537788 (update SRC_URI), raises EAPI and uses distutils-r1 (for the r2 versions). This is my first try using distutils-r1 - i hope it's correct. I've created two r2 for both versions. However, since 0.3.1 is testing anyway, i think we could drop the r1 directly.

[1]https://bugs.gentoo.org/show_bug.cgi?id=537788


net-wireless/python-wifi/python-wifi-0.3.1-r2.ebuild
```diff
@@ -0,0 +1,37 @@
+# Copyright 1999-2012 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: /var/cvsroot/gentoo-x86/net-wireless/python-wifi/python-wifi-0.3.1-r1.ebuild,v 1.3 2012/02/25 11:53:52 patrick Exp $
+
+EAPI=5
+
+PYTHON_COMPAT=( python2_7 )
+DISTUTILS_SINGLE_IMPL=1
```
jlec: Why this?

mm1ke: Hmm, i think i've wrongly interpret SUPPORT_PYTHON_ABIS="1" from the old ebuild. I'll gonna remove DISTUTILS_SINGLE_IMPL.

mgorny:
Is the project alive or used by anything?

mm1ke:
The project doesn't seem to be much alive - last version is from 12/2009, however sys-apps/dstat uses this package via it's wifi use-flag.

net-wireless/python-wifi/python-wifi-0.3.1-r2.ebuild
```diff
@@ -0,0 +1,32 @@
...
+
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+LICENSE="LGPL-2.1 examples? ( GPL-2 )"
+IUSE="examples"
+
+RDEPEND=""
+DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]"
+
+DOCS="docs/AUTHORS docs/BUGS docs/DEVEL.txt docs/TODO"
```
mgorny: Please turn this into an array.

net-wireless/python-wifi/python-wifi-0.3.1-r2.ebuild
```diff
@@ -0,0 +1,32 @@
...
+
+src_install() {
+	distutils-r1_src_install
+	if use examples; then
+		insinto /usr/share/${P}/
+		doins -r examples
```
mgorny: `dodoc -r examples` (without `insinto`) does the same in EAPI 5.

net-wireless/python-wifi/python-wifi-0.3.1-r2.ebuild
```diff
@@ -0,0 +1,29 @@
...
+IUSE="examples"
+
+RDEPEND=""
+DEPEND="dev-python/setuptools[${PYTHON_USEDEP}]"
+
+DOCS=( docs/AUTHORS docs/BUGS docs/DEVEL.txt docs/TODO )
+
+src_install() {
+	distutils-r1_src_install
+	use examples && doins -r examples
```
mgorny: Err, this is wrong. You can't `doins` something without using `insinto` first. And you likely want `dodoc -r` here anyway, since that's what puts files in `/usr/share/doc/${PF}` :).

mgorny:
Ok, I see someone else committed a screwed up -r2, so I've committed your ebuild as -r3. I've also changed it a bit not to install manpages for examples with USE=-examples. Thanks a lot and sorry for the delay.

PR #67, opened 2015-03-09
boot0-10.1: fix compile fails with llvm-3.5.1, bug 542676.

nigoro:
Details please see https://bugs.gentoo.org/show_bug.cgi?id=542676


mgorny:
@nigoro, can I have a bit of your attention on https://bugs.gentoo.org/show_bug.cgi?id=539858#c5, please? I'm still waiting for reply to merge that old PR.

sys-freebsd/boot0/boot0-10.1.ebuild
```diff
@@ -62,6 +62,15 @@ src_compile() {
 		cd "${S}/libstand32" || die
 		freebsd_src_compile
 	fi
+
+	# bug542676
+	if [[ $(tc-getCC) == *clang* ]]; then
+		cd "${S}/i386/btx" || die
+		freebsd_src_compile
+		cd "${S}/i386/boot2" || die
+		CC=gcc freebsd_src_compile
```
mgorny: Hmm, no, you shouldn't ever call `gcc` like this. `${CHOST}-gcc` at least.

mgorny:
Merged, thanks.

PR #70, opened 2015-03-15
freebsd-sources-10.1-r1: Added support upgrades from 9.x.

nigoro:
I was confirmed that a problem with upgrade.
Sorry to trouble you, please merge the patch.

Thanks in advance.


mgorny:
Merged.

PR #71, opened 2015-03-15
sys-devel/binutils-2.25: added ~amd64-fbsd keywords, bug 499212.

nigoro:
This PR is KEYWORDREQ.

https://bugs.gentoo.org/show_bug.cgi?id=499212


mgorny:
Merged, thanks.

PR #74, opened 2015-03-23
xfce-base/exo: version bump

jgeboski:


mgorny:
Thanks for the PR. However, I'm going to upgrade exo myself since I'm adding a few more changes.

PR #75, opened 2015-03-28
Unbreak deps

mgorny:
A few random changes to unbreak the breakage people committed when I wasn't looking…


mgorny:
Emboss causes more issues, kgcc64 is so broken repoman doesn't accept it… why am I wasting my time here?!

PR #77, opened 2015-04-02
sys-firmware/iwl7265-ucode-0.10.13.23: Install iwlwifi-7265D-10.ucode

marbre:
Fix for [Bug 545366](https://bugs.gentoo.org/show_bug.cgi?id=545366).


prometheanfire:
I'm actually going to remove this as I think we should be using linux-firmware from now on

marbre:
Well, if the package is removed from tree okay... Otherwise I would like to see the fix, as it prevents other users running into trouble like me with my new lenovo ;)

prometheanfire:
ya, I'll mask it tonight

marbre:
Closing, as iwl7625-ucode is masked now.

PR #81, opened 2015-04-05
[dev-libs/libdbusmenu] set SLOT=0 and use gtk3 USE instead of gtk

mgorny:
this step makes it easier to move all packages to the new scheme

v2: add revbump

updated version of @karolherbst's patch.


mgorny:
Merged.

PR #84, opened 2015-04-07
BUmp != Bump

jlec:
Signed-off-by: Justin Lecher jlec@gentoo.org


mgorny:
What's the point? Looks like a huge mess just to sweep your typo under the carpet :P.

jlec:
The point of the PR was validation. Which failed. Any chance to fix Travis?

The point of the change was making those guys happy who have problems with these type of errors.

mgorny:
Can't really. Travis can't succeed if the original repo tree contained an error. That's why I shout that loudly when someone introduces dep failures.

PR #92, opened 2015-04-10
Bump nodejs dependency on on libuv.

MJDSys:
nodejs 0.12.2 requires libuv 1.4.2, otherwise many packages fail
with an error similar to:
net.js:164
        throw errnoException(err, 'setBlocking');
              ^
Error: setBlocking ENOSYS

Fixes bug 545618


vanthome:
Please merge

mgorny:
Welcome and thanks a lot for the patch! Since this is a runtime dependency change, and the issue occured after installing nodejs this requires a revision bump. But I'll fix that on my end :).

mgorny:
```
+*nodejs-0.12.2-r1 (11 Apr 2015)
+
+  11 Apr 2015; Michał Górny <mgorny@gentoo.org> +nodejs-0.12.2-r1.ebuild,
+  -nodejs-0.12.2.ebuild:
+  Bump dependency on libuv due to runtime failures, bug #545618.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/92 by Matthew
+  Dawson.
```

PR #95, opened 2015-04-14
net-libs/iojs-1.7.1 version bump

jbergstroem:
Version bump as well as a few improvements. I will add commits as the review progresses and rebase once I get enough LGTM's.

(Intended, see todo) changes from 1.6.3-r1:
- updated metadata
- sort inherit
- use pax-mark on mksnapshot to avoid compile failure on hardened kernels
- support USE=debug which installs a debug version of iojs (otherwise only available in src as iojs_g)
- custom src_install to better handle which files goes where

This ebuild is also meant to land as nodejs-0.12.x once properly reviewed and landed. It should fix all open bugs of nodejs. I sincerely apologise to all nodejs users for leaving the ebuild out in the cold for a while - but my primary interests has lived with iojs for a while.

Open todo's:
- [x] verify that hardened kernel builds properly
- [x] investigate whether paxmark.sh from the elfix package is required or if pax-mark from pax-utils is enough
- [x] make USE=debug _only_ builds debug-related stuff
- [x] don't use tools/install.py for src_install since it doesn't support debug


mgorny:
For the future: please state explicitly that you're the proxied maintainer, so I wouldn't unnecessarily check who to ping about the ebuild :).

jbergstroem:
@mgorny apologies. I'm currently awaiting feedback from hardened-people (while downloading pentoo myself)

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+LICENSE="Apache-1.1 Apache-2.0 BSD BSD-2 MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86 ~x64-macos"
+IUSE="bundled-libs debug icu +npm snapshot +ssl"
+
+RDEPEND="icu? ( dev-libs/icu )
+	!bundled-libs? (
+		>=net-libs/http-parser-2.3
+		>=dev-libs/libuv-1.4.2
+		>=dev-libs/openssl-1.0.1m[-bindist]
```
mgorny: Any sane reason to bundle those three? If it's just 'upstream has configure flags for it', then kill the flag and always use system libs. We allow bundling/static only in special cases (like to solve circular deps or when upstream is going to shout loud at people).

jbergstroem: Yes. One reason is that certain use-cases (containers, et al) benefit from a simpler binary. As of next version of iojs, it even has a fully static (libc included) representation. I felt that exposing this would be beneficial to the user, especially in gentoo.

mgorny: No, it's not beneficial. Nobody's going to track which packages randomly bundle vulnerable libraries. Security is our primary concern, convenience comes later.

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+	tc-export CC CXX PKG_CONFIG
+	export V=1 # Verbose build
+	export BUILDTYPE=Release
+
+	# fix compilation on Darwin
+	# http://code.google.com/p/gyp/issues/detail?id=260
+	sed -i -e "/append('-arch/d" tools/gyp/pylib/gyp/xcode_emulation.py || die
+
+	# make sure we use python2.* while using gyp
+	sed -i -e "s/python/python2/" deps/npm/node_modules/node-gyp/gyp/gyp || die
```
mgorny: No need for that, `python` and `python2` are both wrapped by python-any-r1.

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+	sed -i -e "s|\"lib\"|\"${LIBDIR}\"|" deps/npm/lib/npm.js || die
+
+	epatch "${FILESDIR}"/${PN}-1.2.0-pkgconfig.patch
+
+	# Avoid a test that I've only been able to reproduce from emerge. It doesnt
+	# seem sandbox related either (invoking it from a sandbox works fine).
+	# The issue is that no stdin handle is openened when asked for one.
+	# It doesn't really belong upstream , so it'll just be removed until someone
+	# with more gentoo-knowledge than me (jbergstroem) figures it out.
+	rm test/parallel/test-stdout-close-unref.js
```
mgorny: `|| die`

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+
+	# debug builds. change install path, remove optimisations and override buildtype
+	if use debug; then
+		sed -i -e "s|out/Release/|out/Debug/|g" tools/install.py
+		filter-flags -O*
```
mgorny: Don't do that. `CFLAGS` are meant to be independent of `USE=debug`. One needs to explicitly specify `-O -g` options he needs.

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+	local myarch=""
+	use bundled-libs || myconf+=( --shared-openssl --shared-libuv --shared-http-parser --shared-zlib )
+	use npm || myconf+=( --without-npm )
+	use icu && myconf+=( --with-intl=system-icu )
+	use snapshot && myconf+=( --with-snapshot )
+	use ssl || myconf+=( --without-ssl )
+	use debug && myconf+=( --debug )
+
+	case ${ARCH} in
+	x86) myarch="ia32";;
```
mgorny: Indent this, please. Also, these days you can rely on `${ABI}` directly (without arch conditionals), it's set in all profiles. In fact, I'm not sure if `${ARCH}` is actually safe (vs `use x86` etc.).

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: Since I am part of upstream I think I'm in a position to do something about that, hence the benefit. Also, the flag needs to be enabled which makes me assume the user knows what they're doing. We have a few issues (test failures) as a result of Gentoo having different versions of dependencies. I'm working on improving the test suite to have better resilience for said tests (openssl 1.0.2 and http_parser currently makes two tests if you're on 1.0.2 resp 1.4.x).

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: Did not know. Thanks!

Edit: actually, this needs to stay. These scripts are installed and needs to reference python2 at all times.

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: Just felt weird to have the build generate `-O0` and my own CFLAGS override with `-O2 ...`, but ok.

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+src_compile() {
+	emake -C out mksnapshot
+	pax-mark m "out/${BUILDTYPE}/mksnapshot"
+	emake -C out
+}
+
+src_install() {
+	local LIBDIR="${ED}/usr/$(get_libdir)"
+	emake install DESTDIR="${ED}"
+	use npm && dohtml -r "${LIBDIR}"/node_modules/npm/html/*
```
mgorny: Please use `dodoc -r "${LIBDIR}"/node_modules/npm/html` instead, dohtml is deprecated.

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+	rm -rf "${LIBDIR}"/node_modules/npm/{doc,html}
```
mgorny: `|| die`

net-libs/iojs/iojs-1.6.4.ebuild
```diff
@@ -0,0 +1,129 @@
...
+	find "${LIBDIR}"/node_modules -type f -name "LICENSE" -delete
+
+	pax-mark -m "${ED}"/usr/bin/iojs
+}
+
+src_test() {
+	out/${BUILDTYPE}/cctest || die
+	"${PYTHON}" tools/test.py --mode=${BUILDTYPE,,} -J message parallel sequential || die
+}
```
mgorny: Missing newline at EOF, reconfigure your editor.

net-libs/iojs/iojs-1.6.4.ebuild
mgorny: Please prove me that the Gentoo Security team lists random packages bundling openssl in openssl GLSAs.

net-libs/iojs/iojs-1.6.4.ebuild
mgorny: Generally the idea is that `USE=debug` toggles asserts, extra code paths and the likes.

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: quick question while at this line -- is it ok to use bash4isms (see build type line 128)?

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: I can't argue with that, but wouldn't security team then mark all iojs packages which bundles an older openssl with bundled-libs as vulnerable?

Since `bundled-libs` is [already in tree](https://github.com/gentoo/gentoo-portage-rsync-mirror/blob/master/net-libs/iojs/iojs-1.6.3-r1.ebuild#L24) - shouldn't we think twice before removing it at least?

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: Ok.

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: Thanks, fixed.

jbergstroem:
@mgorny just pushed a new version. Still need to come to a resolution regarding `bundled-libs` and possibly the bash4ism.

net-libs/iojs/iojs-1.6.4.ebuild
mgorny: Not before EAPI 6. Which meas no.

jbergstroem:
Quick update: Vadim A. Misbakh-Soloviov [just reported](https://bugs.gentoo.org/show_bug.cgi?id=540766) that hardened now builds this properly. Also, I've done some reading and gotten help from #gentoo-hardened as well as from blueness@g.o -- we won't be needing paxmark.sh. If there are no more outstanding comments, I'll rebase.

net-libs/iojs/iojs-1.6.4.ebuild
mgorny: If they are installed, then you aren't allowed to use python-any-r1. Use python-single-r1 and python_fix_shebang then.

net-libs/iojs/iojs-1.6.4.ebuild
mgorny: Security team can't keep a huge list of 'who bundles what in which version'. This simply is too much work.

Just because others screwed up hard, doesn't mean we should let things get even worse. I'm just one person, I can't watch everyone all the time :P.

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: This is not fixing shebangs though, it's replacing calls within a script.

net-libs/iojs/iojs-1.6.4.ebuild
jbergstroem: ok, fixing.

net-libs/iojs/iojs-1.6.4.ebuild
mgorny: Ok. Replace them with `${EPYTHON}` then :).

## 

Best regards,
Michał Górny

jbergstroem: ok. fixing now.

mgorny:
I guess this is the point when you can rebase for final review :).

jbergstroem:
@mgorny done. 

Edit: minor spelling issues in commit logs -- feel free to revise logs.

net-libs/iojs/files/iojs-1.7.1-pkgconfig.patch
```diff
@@ -0,0 +1,48 @@
...
+index 67440a2..2a3b05f 100755
+--- configure
++++ configure
+@@ -326,12 +326,13 @@ def b(value):
+ 
+ 
+ def pkg_config(pkg):
+-  cmd = os.popen('pkg-config --libs %s' % pkg, 'r')
++  pkg_config = os.environ.get('PKG_CONFIG', 'pkg-config')
++  cmd = os.popen(pkg_config + ' --libs %s' % pkg, 'r')
```
mgorny, line 11: Are you upstreaming this? :)

net-libs/iojs/iojs-1.7.1.ebuild
```diff
@@ -0,0 +1,127 @@
...
+HOMEPAGE="http://iojs.org/"
+SRC_URI="http://iojs.org/dist/${MY_PV}/${MY_P}.tar.xz"
+
+LICENSE="Apache-1.1 Apache-2.0 BSD BSD-2 MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~arm ~x86 ~x64-macos"
+IUSE="bundled-libs debug icu +npm snapshot +ssl"
+
+RDEPEND="icu? ( dev-libs/icu )
+	!bundled-libs? (
```
mgorny, line 25: Didn't I just tell you this isn't going to fly?

net-libs/iojs/iojs-1.7.1.ebuild
```diff
@@ -0,0 +1,127 @@
...
+		>=net-libs/http-parser-2.3
+		>=dev-libs/libuv-1.4.2
+		>=dev-libs/openssl-1.0.1m[-bindist]
+	)"
+DEPEND="${RDEPEND}
+	${PYTHON_DEPS}
```
mgorny: If scripts use it at runtime, this belongs in RDEPEND.

net-libs/iojs/iojs-1.7.1.ebuild
```diff
@@ -0,0 +1,127 @@
...
+	!!net-libs/nodejs"
+S="${WORKDIR}/${MY_P}"
+
```
mgorny, line 34: `REQUIRED_USE=${PYTHON_REQUIRED_USE}`

net-libs/iojs/iojs-1.7.1.ebuild
```diff
@@ -0,0 +1,127 @@
...
+	# Avoid a test that I've only been able to reproduce from emerge. It doesnt
+	# seem sandbox related either (invoking it from a sandbox works fine).
+	# The issue is that no stdin handle is openened when asked for one.
+	# It doesn't really belong upstream , so it'll just be removed until someone
+	# with more gentoo-knowledge than me (jbergstroem) figures it out.
+	rm test/parallel/test-stdout-close-unref.js || die
+
+	# fix upstream bug regarding shared build logic (my own fault, jbergstroem)
+	sed -i -e "s/== True/== 'true'/g" configure || die
```
mgorny, line 73: I guess the above patch only fixes it partially?

net-libs/iojs/iojs-1.7.1.ebuild
jbergstroem, line 25: I gave you feedback you never replied to - as well as suggesting a mitigation path forward on IRC.

net-libs/iojs/iojs-1.7.1.ebuild
jbergstroem, line 73: Pretty much. See above comment to the proper upstream patch which also makes us drop our pkg-config patch. The only fallout is libuv.

net-libs/iojs/iojs-1.7.1.ebuild
jbergstroem, line 34: Ok.

net-libs/iojs/files/iojs-1.7.1-pkgconfig.patch
jbergstroem, line 11: Yes, [and more](https://github.com/iojs/io.js/pull/1429).

net-libs/iojs/iojs-1.7.1.ebuild
mgorny, line 25: Sorry, must've missed it when you changed the code and github ate old comments.

net-libs/iojs/iojs-1.7.1.ebuild
jbergstroem: Correct. Thanks for spotting.

mgorny:
Ready to merge? No need to rebase, it'll get squashed when applying.

jbergstroem:
I'm all good if you are :)

mgorny:
```
+*iojs-1.7.1 (16 Apr 2015)
+
+  16 Apr 2015; Michał Górny <mgorny@gentoo.org>
+  +files/iojs-1.7.1-pkgconfig.patch, +iojs-1.7.1.ebuild, metadata.xml:
+  Version bump. https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/95 by
+  jbergstroem.

```

PR #96, opened 2015-04-15
Add spacefm-1.0.0.

konsolebox:
spacefm-1.0.0.ebuild is copied from spacefm-0.9.4.ebuild and is modified
to also include an optional use flag disable-red-menu-bar to keep
SpaceFM from setting the background color of the menu bar to red when
SpaceFM is running with EUID 0.

The file spacefm-1.0.0-disable-red-menu-bar.patch is used for the
patching.

spacefm-1.0.0.ebuild also now requires media-video/ffmpegthumbnailer.


mgorny:
The menu bar thingie does not belong in source, and we're certainly not going to apply such a patch unless upstream does. And I doubt they will, unless you make it configurable via some settings menu.

konsolebox:
I don't get it. What do you mean by "it does not belong in source"? Isn't it enough that we allow the user to decide if would use the disable-red-menu-bar flag? What critical issues would we get from here? Is it just because we promote what others would believe as bad practice? Note that there are not many file managers that do this - and perhaps there's only one.

mgorny:
The issue is a that we diverge from upstream, and create extra work for anyone who would like to maintain this package in the future. For example, next version the patch doesn't apply. Now either developer has to work on updating it, wait for someone else to do it or drop it and get complaints that something got dropped. If you have problems with basic principles behind Gentoo, then please discuss them on the mailing lists.

PR #99, opened 2015-04-16
Drop broken oslo-sphinx

mgorny:
Due to dep removal


prometheanfire:
more info?

prometheanfire:
we can remove, though 2.2.0-r1 seems odd It should be fine (pbr dep should probably be >=0.8.0,<1.0.0)

mgorny:
Someone removed old pbr.

PR #101, opened 2015-04-16
[proxy] net-firewall/shorewall-4.6.8.1 version bump

Whissi:
This PR bumps net-firewall/shorewall to v4.6.8.1 and removes old versions.


Whissi:
I updated ChangeLog file; now I read in repository's README that you don't like that. Sorry.
Should I undo or will you take care?

idella:
It's trivial. just leave end I / we shall adjust

mgorny:
It's not that we don't like it, it just doesn't really matter. The merging script just discards the ChangeLog changes.

net-firewall/shorewall/files/4.6/shorewall-lite.systemd-r1
```diff
@@ -0,0 +1,20 @@
...
+Description=Shorewall IPv4 firewall lite
+Documentation=man:shorewall-lite(8) http://www.shorewall.net/Documentation_Index.html
+Wants=network-online.target
+After=network-online.target
+Conflicts=iptables.service firewalld.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+EnvironmentFile=/etc/conf.d/shorewall-lite
```
mgorny, line 14: This is against the [policy](https://wiki.gentoo.org/wiki/Project:Systemd/Ebuild_policy#Unit_file_guidelines) (pt 4), and I'm pretty sure repoman complains for that (you know you ought to `repoman full`, right? :)).

net-firewall/shorewall/files/4.6/shorewall-lite.systemd-r1
```diff
@@ -0,0 +1,20 @@
+#
+#	The Shoreline Firewall Lite (Shorewall-Lite) Packet Filtering Firewall - V4.6
+#
+[Unit]
+Description=Shorewall IPv4 firewall lite
+Documentation=man:shorewall-lite(8) http://www.shorewall.net/Documentation_Index.html
+Wants=network-online.target
+After=network-online.target
+Conflicts=iptables.service firewalld.service
```
mgorny: In Gentoo, `iptables.service` is just a wrapper. Look at how we do firewall units. If shorewall's not using a daemon, you likely want to copy that concept.

Whissi:
@mgorny Sorry, I don't understand your [comment](https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/101/#discussion_r28645650) regarding the wrapper. No, shorewall is not a daemon (it is a combination of Shell and Perl scripts) but I don't understand what's wrong with the quoted code. What's your point?

Regarding your second comment: Are you asking me to create _/etc/systemd/system/shorewall.service.d/gentoo.conf_ where I define `OPTIONS`, `STARTOPTIONS` (i.e. `Environment="OPTIONS=-vvv"`), currently used in the service files and read from the EnvironmentFile?

Please be aware that I don't use systemd. I don't feel comfortable in changing anything I cannot test...

mgorny:
I was pointing out that the conflict on iptables doesn't actually match iptables. And for the second comment, yes, kind of that.

Let's try to CC some help: @floppym, @jcallen, @rich0, @blueness, @tamiko, @jauhien.

Whissi:
Ah, now I understand. So we need to add _iptables-restore.service_ to _Conflicts_, too. I'll update the PR soon.

I can also make the changes for systemd to get rid of `EnvironmentFile` but can you please show me how I pass multiple vars with values containing spaces to systemd? I.e. I need to pass

``` Bash
PRODUCTS="shorewall shorewall6"
STARTOPTIONS="-vvv"
```

How must the _/etc/systemd/system/shorewall-init.service.d/00gentoo.conf_ file look like?

Edit: OK, according the man page multiple _Environment_ statements are allowed. Are values with spaces a problem or would something like

```
Environment="PRODUCTS=shorewall shorewall6"
Environment="STARTOPTIONS=-vvv"
```

work?

mgorny:
I don't think `PRODUCTS` is actually used anywhere. And you shouldn't export `STARTOPTIONS`, just let people redefine ExecStart= if they need any.

Whissi:
Wrong, `PRODUCTS` is used, see `/usr/sbin/shorewall-init`:

``` BASH
# Initialize the firewall
shorewall_start () {
  local PRODUCT
  local STATEDIR

  echo -n "Initializing \"Shorewall-based firewalls\": "
  for PRODUCT in $PRODUCTS; do
      setstatedir

      if [ -x ${STATEDIR}/firewall ]; then
          #
          # Run in a sub-shell to avoid name collisions
          #
          (
              if ! ${STATEDIR}/firewall status > /dev/null 2>&1; then
                  ${STATEDIR}/firewall ${OPTIONS} stop || exit 1
              else
                  exit 1
              fi
          )
      else
          echo ERROR:  ${STATEDIR}/firewall does not exist or is not executable!
          exit 1
      fi
  done

  return 0
}
[...]
case "$1" in
  start)
     shorewall_start
     ;;
[...]
```

mgorny:
Now I don't get it. So you create two services that both start both firewalls depending on `PRODUCTS`? Anyway, having multiple products in a single unit is an ugly openrc-ism.

Whissi:
First, this is upstream's design. That's no Gentoo-specific stuff.

Shorewall consists of several components:
- shorewall (IPv4 Firewall)
- shorewall6 (based on shorewall, adds IPv6 support)
- shorewall-lite (same like shorewall without the compiler)
- shorewall6-lite (same like shorewall6 without the compiler)
- shorewall-init (will initialize the _products_ above at boot time prior to bringing up the network or reacts on changing interfaces)

Now, _shorewall-init_ needs to know which _product_ (a shorewall term) should be initialized at boot. Let's treat _shorewall-init_ as a normal service, _/etc/conf.d/shorewall-init_ is its configuration file.

From my understanding "you" (the systemd folk) don't loke _/etc/conf.d_ because these files could be shell scripts because this is allowed in OpenRC's world. So why not just have an own config file just for systemd which will be used in _EnvironmentFile_? Should solve this problem.

Is something like

```
ExecStart=PRODUCTS="shorewall shorewall6" /usr/sbin/shorewall-init start
```

allowed?

mgorny:
But you've got 5 systemd units… so if init is supposed to take care of all of them, what do the remaining units do?

systemd folk doesn't like when services started are unpredictable. That's why we do no magic variables `START_THIS` but instead expect people to enable components via enabling the respective units, i.e.:

```
systemctl enable iptables ip6tables
```

not:

```
echo TABLES="iptables ip6tables arptables" > somewhere
systemctl enable all-tables
```

Whissi:
You have 5 systemd units because these are 5 different things.
On a typical system you will only use shorewall[6]? or shorewall[6]?-lite (both with or without shorewall-init). So you end up with 2 or 3 units.

It is the same like when you are using iptables: You have _iptables-restore_ (=shorewall or shorewall-lite) and _ip6tables-restore_ (shorewall6 or shorewall6-lite). Shorewall-init is just an additional application (script) which makes sure that while booting your system is in a secure state (=nothing can connect between the time an interface was started and the firewall was started).

Maybe it would be easier to understand if we would have _shorewall-init_ **and** _shorewall6-init_... but upstream created **one** shorewall-init application (which is more like a wrapper so it makes sense to have only **one** wrapper which can be used for IPv4 **and** IPv6) which is able to work with _shorewall_ **and** _shorewall6_ (or _shorewall-lite_ **and** _shorewall6-lite_).

...and this wrapper needs to know if it should handle _shorewall_ **and/or** _shorewall6_ (or in theory you could have _shorewall_ (the real IPv4 firewall with local configuration and compiler) and _shorewall6-lite_ (only the wrapper, the configuration must be created on another system with _shorewall6_)...

mgorny:
Well, it all sounds like a lot of scary magic to me. I'd normally suggest splitting shorewall-init, assuming it could be run twice with different products. But if it's no regression, I don't care that much.

So please let me know if there's anything else you'd like to change or whether I should just merge it.

Whissi:
Because this is just a simple bump (bugfix release, nothing new) and I cannot test systemd I don't want to change anything regarding systemd implementation (I will only update the _Conflicts_ line to match Gentoo's iptables services).

If you want me to get rid of _/etc/conf.d/_ usage I would create _/etc/systemd/system/shorewall__.service.d/00gentoo.conf_ for each service. For _shorewall-init_ it would look like

```
/etc/systemd/system/shorewall-init.service.d/00gentoo.conf:

[Service]
Environment="PRODUCTS=shorewall shorewall6"
Environment="STARTOPTIONS=-vvv"
```

```
/usr/lib/systemd/system/shorewall-init.service:

[Unit]
Description=shorewall-init
Documentation=http://www.shorewall.net/Shorewall-init.html
Before=network-pre.target
Wants=network-pre.target
Conflicts=iptables.service firewalld.service iptables-restore.service ip6tables-restore.service

[Service]
Type=oneshot
RemainAfterExit=yes
StandardOutput=syslog
ExecStart=/usr/sbin/shorewall-init $STARTOPTION start
ExecStop=/usr/sbin/shorewall-init $STARTOPTION stop

[Install]
WantedBy=basic.target

```

Would that work?

Please tell if I should do that or only update the _Conflocts_ line and keep the rest for the moment.

mgorny:
Keep the rest for now. Hopefully someday you will be able to test the other changes :P.

Whissi:
We'll see ;-)

I updated the PR to include the changes for the _Conflicts_ line. The PR is now ready to be merged.

mgorny:
```
+*shorewall-4.6.8.1 (22 Apr 2015)
+
+  22 Apr 2015; Michał Górny <mgorny@gentoo.org>
+  +files/4.6/shorewall-init.systemd-r1, +files/4.6/shorewall-lite.systemd-r1,
+  +files/4.6/shorewall.systemd-r1, +files/4.6/shorewall6-lite.systemd-r1,
+  +files/4.6/shorewall6.systemd-r1, +shorewall-4.6.8.1.ebuild,
+  -shorewall-4.6.5.3.ebuild, -shorewall-4.6.6.1.ebuild:
+  Version bump, drop old. https://github.com/gentoo/gentoo-portage-rsync-
+  mirror/pull/101 by Whissi.
```

PR #104, opened 2015-04-20
xfce-extra/thunar-archive-plugin: New ebuild fixing #518180 and adding engrampa support

gbraekmans:
Second try!

I've added the EAUTORECONF='yes' line to the ebuild. Is it in the correct section? Do the build dependancies change because I've added this line and we need  xfce4-dev-tools now? Or is this handled in the eclass?

Regards,
Geert


mgorny:
What's this engrampa thing? Is this upstream backport, or some custom patch?

gbraekmans:
It's the file-roller fork from mate. It would be a custom patch I guess.

mgorny:
(CC @perfinion)

I don't think we should apply custom patches like this, unless upstream agrees on it. Is there an upstream bug report about this? Could you open one?

gbraekmans:
Several reports exist upstream. No need to add another ;)

Upstream seems to agree, from what I can find in https://bugzilla.xfce.org/show_bug.cgi?id=10814 It is not really a bug to them (and indeed no source code needs to be changed, just a file added)

perfinion:
http://git.xfce.org/thunar-plugins/thunar-archive-plugin/log/
I dont see any mention of engrampa in upstream's repo. I'd also prefer them to commit it first. when they commit it i'd fine with adding it as a patch until the next release if thats not going to come for a long time.

gbraekmans:
Allright, but the same principle can be applied for the bug (same problem, different executable). So we keep the package "broken" until upstream decides to spend some time cleaning up their bugs?

mgorny:
Well, I'm going to take the engrampa patch, clean it up a bit and add to upstream bug report. Patches often work better than trivial 'do X and Y' explanations :).

What about the other patch? Where does it come from?

gbraekmans:
Same thing but for file-roller. Upstream has already a patch for that in the bugzilla.

gbraekmans:
Allright, should I keep this in my local overlay? Add a mate useflag for engrampa support? Are we going to talk to upstream first? I would really keep the fileroller-patch, because at the moment file-roller support is broken out of the box... Make a decision devs and let's get this request (and bug report) closed.

@mgorny look at  xfce-extra/thunar-archive-plugin/files/fix-file-roller.patch in the diff. The source or the patch was #518180

gbraekmans:
I guess I should be using branches...

mgorny:
Yes, branches would be preferred. I'm sorry but I am currently quite busy at work, and my time for Gentoo is limited. I have already tried to contact upstream, and it seems that this particular plugin is not really maintained ;/. I need a second opinion but I think we're going to merge your patches.

mgorny:
@perfinion, could you clean this up and fix whitespace in patches?

perfinion:
-  19 May 2015; Jason Zaman perfinion@gentoo.org
-  +files/0.3.1-add-engrampa-support.patch, +files/0.3.1-fix-file-roller.patch,
-  +thunar-archive-plugin-0.3.1-r1.ebuild:
-  fix fileroller and engrampa. PR https://github.com/gentoo/gentoo-portage-
-  rsync-mirror/pull/104 thanks to brakarov

Its merged in, I cleaned up the whitespace and a minor issues (the EAUTORECONF only needs to be set once, and must be before the inherit)

looks like i dont have perms to close this PR tho

mgorny:
Thanks a lot!

PR #105, opened 2015-04-20
Added new dependency to CGO usb bindings

likewhoa:
This is in relation to bug 547204 at b.g.o which bug 523990 needs in order to properly compile. I will send another PR and reference it which will contain the ebuild for dev-libs/go-usb


likewhoa:
Commit for dev-libs/go-usb package. Please review.

mgorny:
Please leave comment on the bug when you submit a PR, so the assignee knows about it :P.

dev-libs/go-usb/go-usb-9999.ebuild
```diff
@@ -0,0 +1,43 @@
...
+
+	for i in LICENSE *.go; do
+		cp "${S}/${i}" "${D}/${GO_PN}" || die "Install failed"
+	done
+}
+
+src_test() {
+	go test -ldflags '-extldflags=-fno-PIC' ${GO_PN} || die
+}
+
```
mgorny: Stray empty line.

dev-libs/go-usb/go-usb-9999.ebuild
```diff
@@ -0,0 +1,43 @@
...
+# Tests require a connected mtp device
+RESTRICT="test"
+
+GO_PN="/usr/$(get_libdir)/go/src/github.com/hanwen/usb"
+
+src_install() {
+	dodir "${GO_PN}"
+
+	rm -r "${S}/.git" || die "Failed to remove .git directory"
```
mgorny: Why is this necessary?

dev-libs/go-usb/go-usb-9999.ebuild
```diff
@@ -0,0 +1,43 @@
+# Copyright 1999-2014 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+inherit git-r3 flag-o-matic multilib toolchain-funcs
```
mgorny: Most of them unused :P.

dev-libs/go-usb/go-usb-9999.ebuild
likewhoa: I'll fix this and those below during the week. Sorry for late reply, it seems i wasn't getting these email notifications.

dev-libs/go-usb/go-usb-9999.ebuild
likewhoa: I removed most of that now, do review.

dev-libs/go-usb/go-usb-9999.ebuild
likewhoa: Yea, this was a copy from go-mtpfs but i cleaned it up now.

dev-libs/go-fuse/go-fuse-9999.ebuild
```diff
@@ -46,7 +46,7 @@ done
 
 insinto /usr/lib/go/
 doins -r "${S}/pkg"
-insinto "/usr/lib/go/src/pkg/${GO_PN}/"
+insinto "/usr/lib/go/src/${GO_PN}/"
```
mgorny, line 49: I've pinged go-fuse maintainer about this.

dev-libs/go-usb/go-usb-9999.ebuild
```diff
@@ -0,0 +1,36 @@
...
+KEYWORDS=""
+IUSE=""
+
+DEPEND=">=dev-lang/go-1.4"
+RDEPEND=""
+
+# Tests require a connected mtp device
+RESTRICT="test"
+
+GO_PN="/usr/$(get_libdir)/go/src/github.com/hanwen/usb"
```
mgorny: Are you sure it's supposed to be libdir and not plain `lib`? go-fuse above uses `lib`. This is important for systems where lib!=lib64.

dev-libs/go-usb/go-usb-9999.ebuild
```diff
@@ -0,0 +1,36 @@
...
+
+src_install() {
+	insinto "${GO_PN}"
+
+	for i in LICENSE *.go; do
+	  doins "${i}" || die "Install failed"
```
mgorny: `doins` can take multiple files. No need to loop.

dev-libs/go-usb/go-usb-9999.ebuild
likewhoa: I am not sure actually so perhaps it should be removed.

dev-libs/go-usb/go-usb-9999.ebuild
likewhoa: The problem was that doing 'doins' by itself was putting it into ${D}/${PF} for some reason. I will test again to confirm.

likewhoa:
@ZeroChaos- see my last commit, i removed the added get_libdir and doins loop since it's not even necessary.

ZeroChaos-:
still approved

mgorny:
Ok, checked this and go itself uses /usr/lib.

dev-libs/go-usb/go-usb-9999.ebuild
```diff
@@ -0,0 +1,33 @@
...
+RDEPEND=""
+
+# Tests require a connected mtp device
+RESTRICT="test"
+
+GO_PN="/usr/lib/go/src/github.com/hanwen/usb"
+
+src_install() {
+	insinto "${GO_PN}"
+	doins *.go LICENSE || die "Install failed"
```
mgorny, line 28: die unnecessary in EAPI 5 (with doins).

mgorny:
Ok, I'm going to fix that tiny bug and merge it.

mgorny:
All merged now. Closing.

PR #106, opened 2015-04-21
Version bump: net-libs/iojs-1.8.1

jbergstroem:
- rebaked pkgconfig patch (I will upstream a proper version soon)
- removed sed since it's fixed upstream


mgorny:
I presume it's fine to review just the 'fixup!' stuff. Except for the minor stuff I pointed out, the patch looks fine. Thanks.

jbergstroem:
Yeah -- copied ebuild as you requested.

As for the patch, I simplified just to get it in. If it's ok with you I'll focus on doing a proper version upstream instead.

mgorny:
Sure, that works for me.

```
+*iojs-1.8.1 (22 Apr 2015)
+
+  22 Apr 2015; Michał Górny <mgorny@gentoo.org>
+  +files/iojs-1.8.1-pkgconfig.patch, +iojs-1.8.1.ebuild:
+  Version bump with rebaked pkg-config patch. https://github.com/gentoo/gentoo-
+  portage-rsync-mirror/pull/106 by Johan Bergström.
```

PR #110, opened 2015-04-27
isl: improve Gentoo Prefix compatibility

alyst:
The current autoload-dir patch installs the python files into `/usr/share/`...
For Gentoo Prefix it should be rather `$(EPREFIX)/usr/share`...
The proper way for both normal and prefix installs would be to use `$(prefix)`.


mgorny:
I'm sorry this is taking so long. Toolchain stuff is kinda fragile, and I'm still asking people to review so that we don't break some fancy use case accidentally :).

mgorny:
Merged, thanks.

PR #111, opened 2015-04-29
Bug #547270 Add gtk3 useflag -- for Xfce 4.12 instead of only libxfce4ui

gbraekmans:
It's now possible to build xfce4-meta using only GTK2 libs. I've included the garcon 0.4 ebuild to support just that, 0.5 has a dependency on GTK3.

I've checked the revdeps for gtk3 support and only these 3 (garcon, indicator and pulseaudio plugin) have problems with non-gtk3 enabled build.

### DIFF
--- a/xfce-base/garcon/garcon-0.4.0.ebuild
+++ b/xfce-base/garcon/garcon-0.4.0.ebuild
@@ -16,9 +16,7 @@ IUSE="debug"
 
 RDEPEND=">=dev-libs/glib-2.30:=
 	>=x11-libs/gtk+-2.24:2=
-	>=x11-libs/gtk+-3.14:3=
-	>=xfce-base/libxfce4ui-4.11.1:=[gtk3(+)]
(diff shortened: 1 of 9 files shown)


karllinden:
Oh, sorry. I did not see this until I had submitted PR #112.

karllinden:
pragha, parole and appfinder need libxfce4ui-2, too, which is what the --enable-gtk3 use flag of libxfce4ui triggers.

Additionally I sent a mail to Landry Breuil at the garcon upstream with a patch to optionally build against gtk3, so hopefully garcon-0.4 should not be needed. Instead garcon-0.5.x can be bumped when the patch is included or the patch can be backported. My fork of garcon containing the fixes can be found at https://gitlab.com/lilrc/garcon

gbraekmans:
Great. It seems you did a better job! ;) I'll close this one. I've just checked the xfce- categories. Could you add my patches for xfce-panel and app finder?

PR #114, opened 2015-05-05
Add multilib support to media-libs/phonon

MJDSys:
This PR adds multilib support to media-libs/phonon, based upon how libdbusmenu-qt implemented it.

This also updates the SRC_URI of the -9999 ebuild to match the 4.8.6 ebuild.


mgorny:
@pesa, @kensington, …?

Pesa:
this is a kde package so it should be handled by them (but I'm happy to review patches)

Pesa:
`[${MULTILIB_USEDEP}]` deps are missing everywhere.

Also how do we want to handle the fact that qt5 is not multilib yet?

kensington:
qt5 handling seems to be the same as in dev-libs/libdbusmenu-qt which is OK, we can update when later when necessary.

I wonder how useful this is without backend plugins also being multilib though.

MJDSys:
@kensington Primarilly, it allows Qt to be compiled in multilib with a KDE desktop installed.  If it weren't for that, I wouldn't even have noticed.

I believe GStreamer is multilib now(?), so I can try playing with phonon-gstreamer later.  But mainly I just need my copy of Qt to compile.

@Pesa Ack on the deps issue, I'll work on that this weekend.

Other thought:  multilib Qt is missing the ${MULTLIB_USEDEP} part on media-libs/phonon, which allowed portage to resolve the dependencies.  Since this PR adds multilib support, I guess I should update the Qt ebuilds as well.  Should I add that to this PR?  Or open a separate one here, or against the Qt repository?

kensington:
Where does Qt require phonon?

MJDSys:
qtdemo failed to compile for me because of this.  A quick grep suggests it's the only one affected, except maybe designer.  I didn't investiage further, as this seemed the simplest solution to fix it.

Pesa:
My point is: building with `USE=qt5` and `ABI_X86=32` is **not** going to produce a 32-bit `libphonon4qt5.so`, is there a way to make it clearer?

I agree with @MJDSys that plugins are not strictly necessary and can be converted later.

About adding `[${MULTILIB_USEDEP}]` on media-libs/phonon deps, just include it here so we don't forget. But in any case this PR is not going to be merged directly, so I don't particularly care which way you go.

mgorny:
No, there's no way to make it cleaner. I just opened 'reminder' bugs stating which USE flags were left inactive.

mgorny:
Anyway, is there anything I should do here or will you handle it?

kensington:
If the multilib stuff looks otherwise fine, I think it's OK to merge with the multilib usedeps added.

mgorny:
I don't see anything obviously wrong though Qt team should know better about their multilib stuff :P.

mgorny:
@MJDSys, are you still there? @Pesa had some comments for you.

MJDSys:
Sorry, I got caught up with some other things.  I think this update covers all the raised issues.

MJDSys:
Fixed -9999 ebuild.

media-libs/phonon/phonon-9999.ebuild
```diff
@@ -44,10 +44,10 @@ RDEPEND="
 		designer? ( dev-qt/designer:5 )
 	)
 	pulseaudio? (
-		dev-libs/glib:2
-		>=media-sound/pulseaudio-0.9.21[glib]
+		dev-libs/glib:2[${MULTILIB_USEDEP}]
+		>=media-sound/pulseaudio-0.9.21[glib,${MULTILIB_USEDEP}]
 	)
-	zeitgeist? ( dev-libs/libqzeitgeist )
+	zeitgeist? ( dev-libs/libqzeitgeist[${MULTILIB_USEDEP}] )
```
mgorny, line 50: libqzeitgeist is not multilib. You'd need to convert it first :).

media-libs/phonon/phonon-9999.ebuild
```diff
@@ -44,10 +44,10 @@ RDEPEND="
...
 "
 DEPEND="${RDEPEND}
```
mgorny, line 52: virtual/pkgconfig needs MULTILIB_USEDEP too.

media-libs/phonon/phonon-9999.ebuild
MJDSys, line 50: Can zeitgeist support just block multilib instead?  That use flag is masked for this package on amd64 anyways, and other arches just block it.  It looks like converting that will be a lot of work for not so much benefit.
If unmasking the useflag is wanted, I can see about working on its dependencies.  Would that work?

Pesa, line 50: I'd say it's better to just remove `[${MULTILIB_USEDEP}]` and enable zeitgeist only for the native ABI until the dep gets converted. This is what I did e.g. for the rarely used backends of `dev-qt/qtsql:4`.

mgorny:
```
+*phonon-4.8.3-r1 (02 Jun 2015)
+
+  02 Jun 2015; Michał Górny <mgorny@gentoo.org> +phonon-4.8.3-r1.ebuild,
+  phonon-9999.ebuild:
+  Enable multilib support and sync the live ebuild. https://github.com/gentoo
+  /gentoo-portage-rsync-mirror/pull/114 by MJDSys with some minor changes.
```

PR #115, opened 2015-05-06
Version bump: net-libs/libuv-1.5.0

jbergstroem:
Rename works fine for me.

/R=@hasufell?


dev-libs/libuv/libuv-1.5.0.ebuild
```diff
@@ -0,0 +1,48 @@
...
+SRC_URI="https://github.com/libuv/libuv/archive/v${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="BSD BSD-2 ISC MIT"
+SLOT="0/1"
+KEYWORDS="~amd64 ~arm ~ppc ~ppc64 ~sparc ~x86 ~amd64-linux ~x86-linux"
+IUSE="static-libs"
+
+DEPEND="
+	sys-devel/libtool
+	virtual/pkgconfig
```
mgorny: MULTILIB_USEDEP here, I think.

jbergstroem: Could you elaborate on this?

jbergstroem: I found no other reference to a combination of either pkgconfig or libel with MULTILIB_USEDEP which his why I was asking.

mgorny: Your grepping skills must be weak :P. Any multilib package using pkg-config should have MULTILIB_USEDEP here, to have `i686-pc-linux-gnu-pkgconfig` etc.

dev-libs/libuv/libuv-1.5.0.ebuild
```diff
@@ -0,0 +1,48 @@
...
+"
+
+src_prepare() {
+	echo "m4_define([UV_EXTRA_AUTOMAKE_FLAGS], [serial-tests])" \
+		> m4/libuv-extra-automake-flags.m4 || die
+
+	sed -i \
+		-e '/CC_CHECK_CFLAGS_APPEND(\[-g\])/d' \
+		configure.ac || die "fixing CFLAGS failed!"
```
mgorny: Doesn't this use some fancy autoconf caching, so you could pass `cv_something_something=no` instead? :) i.e. look how I disable gold linker in systemd.

jbergstroem:
@mgorny any additional feedback?

hasufell:
@mgorny @mgorny @mgorny :dancer: :1234:

mgorny:
```
+*libuv-1.5.0 (18 May 2015)
+
+  18 May 2015; Michał Górny <mgorny@gentoo.org> +libuv-1.5.0.ebuild:
+  Version bump. https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/115
+  by jbergstroem.
```

hasufell:
:100:

dev-libs/libuv/libuv-1.5.0.ebuild
```diff
@@ -0,0 +1,45 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+inherit autotools-utils eutils multilib multilib-minimal
```
hasufell, line 7: Please don't convert my ebuilds to use `autotools-utils`. It even introduced a bug. This has been reverted in `1.7.5`.

PR #116, opened 2015-05-09
[media-sound/teamspeak-client-bin] fix bug #548048

karolherbst:
link to bug: https://bugs.gentoo.org/show_bug.cgi?id=548048

Package-Manager: portage-2.2.18


media-sound/teamspeak-client-bin/teamspeak-client-bin-3.0.16.ebuild
```diff
@@ -20,6 +20,7 @@ IUSE="alsa pulseaudio"
 REQUIRED_USE="|| ( alsa pulseaudio )"
 
 RDEPEND="
+	dev-libs/quazip[qt5]
```
mgorny: Since this is a binary package, it would be nice to depend on a specific subslot (since no other subslot will work). But I don't see a subslot declared in quazip, so I'm going to ask @jlec.

@jlec, can we assume that subslot `:0/0` will have libquazip.so.1, or do you plan to change that to `:0/1`?

jlec: I have no specific plan. All versions in tree have soname 1.0.0. But perhaps it would be beneficial for at least this package to have a subslot declared. Feel free to add it.

karolherbst: I am a bit worried about the either qt4 or qt5 REQUIRED_USE in quazip :/

karolherbst: I was already thinking about adding a system-quazip USE flag in case somebody needs quazip[qt4] installed

mgorny: If someone needs that, then we will have to fix quazip instead.

I've talked to jlec, and please add `:0/0` there.

karolherbst: done

mgorny:
Merged, thanks.

PR #117, opened 2015-05-09
[dev-libs/libappindicator] fix slot dependency of libindicator

karolherbst:
a small typo I saw while depcleaning my system


karolherbst:
as it seems it was fixed now: https://bugs.gentoo.org/show_bug.cgi?id=549064

PR #118, opened 2015-05-11
Improvements to the nodejs 0.12.2 ebuild

jbergstroem:
Most of these changes are adopted from iojs, but since nodejs hasn't evolved the same way as iojs (both build system and src), a lot of stuff is held back. This is intentional from an ebuild point of view as well to treat nodejs 0.x as stable builds forward.

This intends to fix the following bugs:
- https://bugs.gentoo.org/show_bug.cgi?id=542970
- https://bugs.gentoo.org/show_bug.cgi?id=539342
- https://bugs.gentoo.org/show_bug.cgi?id=539314
- https://bugs.gentoo.org/show_bug.cgi?id=540766
- https://bugs.gentoo.org/show_bug.cgi?id=541656
- https://bugs.gentoo.org/show_bug.cgi?id=462703
- https://bugs.gentoo.org/show_bug.cgi?id=540106
- (pending confirm) https://bugs.gentoo.org/show_bug.cgi?id=533176


mschubert:
There are still issues with https://bugs.gentoo.org/show_bug.cgi?id=541656 for prefix, see my comment.

mgorny:
@jbergstroem, it looks that patrick has added -r2 with some random changes. Could you take a look and update the PR?

jbergstroem:
Ok - he committed the first version. Rebased this patchset onto master -- recommending -r3 to fix the prefix install :tada:

jbergstroem:
After that's been cut we should also close all referenced bugs.

mgorny:
https://bugs.gentoo.org/show_bug.cgi?id=539314 is pengupop stablereq :P.

mgorny:
Ok, merged as -r3 and closed all mentioned bugs. Though you may want to check whether 539314 didn't mean something else :P.

PR #120, opened 2015-05-15
dev-db/redis Verbump 3.0.2

jbergstroem:
Simple rename. 

Lets keep this PR open so we can possibly add a few fixes for open bugs, namely related to systemd.

Ping @ultrabug


mgorny:
Saying 'simple rename' is not helpful. Would be better if you told 'I compared configure.ac / CMakeLists.txt' :P.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+	# Copy lua modules into build dir
+	cp "${S}"/deps/lua/src/{fpconv,lua_bit,lua_cjson,lua_cmsgpack,lua_struct,strbuf}.c "${S}"/src || die
+	cp "${S}"/deps/lua/src/{fpconv,strbuf}.h "${S}"/src || die
+	# Append cflag for lua_cjson
+	# https://github.com/antirez/redis/commit/4fdcd213#diff-3ba529ae517f6b57803af0502f52a40bL61
+	append-cflags "-DENABLE_CJSON_GLOBAL"
+
+	# now we will rewrite present Makefiles
+	local makefiles=""
+	for MKF in $(find -name 'Makefile' | cut -b 3-); do
```
mgorny: Nope, not going to like hackery like that. What's the deal with cutting here anyway? Use proper `while read; ...; done < <(...)`. Look at python-utils-r1 how use `find` like that.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+		mv "${MKF}" "${MKF}.in"
```
mgorny: ||die.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+
+	# Copy lua modules into build dir
+	cp "${S}"/deps/lua/src/{fpconv,lua_bit,lua_cjson,lua_cmsgpack,lua_struct,strbuf}.c "${S}"/src || die
+	cp "${S}"/deps/lua/src/{fpconv,strbuf}.h "${S}"/src || die
+	# Append cflag for lua_cjson
+	# https://github.com/antirez/redis/commit/4fdcd213#diff-3ba529ae517f6b57803af0502f52a40bL61
+	append-cflags "-DENABLE_CJSON_GLOBAL"
+
+	# now we will rewrite present Makefiles
+	local makefiles=""
```
mgorny: Make this an array.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+DESCRIPTION="A persistent caching system, key-value and data structures database"
+HOMEPAGE="http://redis.io/"
+SRC_URI="http://download.redis.io/releases/${P}.tar.gz"
+
+LICENSE="BSD"
+KEYWORDS="~amd64 ~hppa ~ppc64 ~x86 ~amd64-linux ~x86-linux ~x86-macos ~x86-solaris"
+IUSE="+jemalloc tcmalloc test"
+SLOT="0"
+
+RDEPEND=">=dev-lang/lua-5.1:*
```
mgorny: It links to lua, so I'd say `:=` would be more correct.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+		configure.ac || die "Sed failed for configure.ac"
+	eautoconf
+}
+
+src_configure() {
+	econf
+
+	# Linenoise can't be built with -std=c99, see https://bugs.gentoo.org/451164
+	# also, don't define ANSI/c99 for lua twice
+	sed -i -e "s:-std=c99::g" deps/linenoise/Makefile deps/Makefile || die
```
mgorny: Why not remove that from `Makefile.in` in src_prepare()?

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+}
+
+src_compile() {
+	tc-export CC AR RANLIB
+
+	local myconf=""
```
mgorny: Array could be cleaner here.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+
+	if use tcmalloc ; then
+		myconf="${myconf} USE_TCMALLOC=yes"
```
mgorny: or at least `+=`.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+		myconf="${myconf} JEMALLOC_SHARED=yes"
+	else
+		myconf="${myconf} MALLOC=yes"
+	fi
+
+	emake ${myconf} V=1 CC="${CC}" AR="${AR} rcu" RANLIB="${RANLIB}"
+}
+
+src_install() {
+	insinto /etc/
```
mgorny: I'd avoid that trailing slash.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+	doins redis.conf sentinel.conf
+	use prefix || fowners redis:redis /etc/{redis,sentinel}.conf
```
mgorny: That's a weird conditional. Anyway, I'd really prefer if you explicitly had IUSE=prefix.

dev-db/redis/redis-3.0.1.ebuild
```diff
@@ -0,0 +1,116 @@
...
+	fperms 0644 /etc/{redis,sentinel}.conf
+
+	newconfd "${FILESDIR}/redis.confd" redis
+	newinitd "${FILESDIR}/redis.initd-4" redis
+
+	systemd_dounit "${FILESDIR}/redis.service"
+	systemd_newtmpfilesd "${FILESDIR}/redis.tmpfiles" redis.conf
+
+	nonfatal dodoc 00-RELEASENOTES BUGS CONTRIBUTING MANIFESTO README
```
mgorny: Nonfatal is wrong here. It's used when you handle failures on their own, not to ignore crap code.

jbergstroem:
FWIW, I didn't imply that 'simple rename' didn't mean work, rather what I did.

dev-db/redis/redis-3.0.1.ebuild
jbergstroem: I found similar stuff in gdb, llvm (clang) and sqlite without having IUSE=prefix. Since we always seem to install them I guess there's no point in checking for prefix before changing ownership?

mgorny:
Sure, just telling it would be nice if you confirmed that you checked stuff :).

dev-db/redis/redis-3.0.1.ebuild
mgorny: Lack of USE=prefix is legitimate due to implicit IUSE stuff. But it's far from clean, and I'd rather kill it completely.

And yes, I have no clue why the ownership is set only conditionally.

jbergstroem:
Note: depends on #136.

jbergstroem:
todo:
- merge #136 
- systemd init/pid debacle
- rewrite the makefile iteration

mrueg:
Looks like it was bumped today, could you please check if all your changes to into this?
*redis-3.0.2 (08 Jul 2015)

  08 Jul 2015; Ultrabug ultrabug@gentoo.org -redis-3.0.0.ebuild,
  +redis-3.0.2.ebuild, +files/redis.service-2:
  version bump fix #551746, change systemd service wrt #537384, drop old

mgorny:
@jbergstroem, where are you? :P

jbergstroem:
@mgorny travelling at the moment - sorry. I'll get onto this as soon as I get a few hours.

mgorny:
1. No ChangeLogs in git.

PR #122, opened 2015-05-17
Version bump: net-libs/iojs-2.0.2

jbergstroem:
Deps on #115.


mgorny:
```
+*iojs-2.0.2 (18 May 2015)
+
+  18 May 2015; Michał Górny <mgorny@gentoo.org> +iojs-2.0.2.ebuild:
+  Version bump. https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/122
+  by jbergstroem.
```

PR #123, opened 2015-05-18
Version bump: dev-libs/libconfig-1.5

jbergstroem:
Rebaked patch since upstream seems to partially (almost) have merged it.


dev-libs/libconfig/libconfig-1.5.ebuild
```diff
@@ -0,0 +1,55 @@
...
+AUTOTOOLS_AUTORECONF="1"
+inherit eutils autotools-multilib
+
+DESCRIPTION="Libconfig is a simple library for manipulating structured configuration files"
+HOMEPAGE="http://www.hyperrealm.com/libconfig/libconfig.html"
+SRC_URI="http://www.hyperrealm.com/${PN}/${P}.tar.gz"
+
+LICENSE="LGPL-2.1"
+SLOT="0"
+KEYWORDS="amd64 ~arm ~mips ppc ppc64 sparc x86 ~x86-linux"
```
mgorny: :P

dev-libs/libconfig/libconfig-1.5.ebuild
```diff
@@ -0,0 +1,55 @@
...
+multilib_src_install_all() {
+	einstalldocs
+	prune_libtool_files
+
+	if use examples; then
+		find examples/ -name "Makefile.*" -delete || die
+		local dir
+		for dir in examples/c examples/c++; do
+			insinto /usr/share/doc/${PF}/${dir}
+			doins ${dir}/*
```
mgorny: docinto; dodoc. In EAPI5 they will be compressed anyway.

mgorny: In fact, `dodoc -r` instead of the whole loop :P.

dev-libs/libconfig/files/libconfig-1.5-out-of-source-build.patch
```diff
@@ -0,0 +1,13 @@
+diff --git tests/Makefile.am tests/Makefile.am
```
mgorny, line 1: I don't see that patch on upstream github!

dev-libs/libconfig/libconfig-1.5.ebuild
```diff
@@ -0,0 +1,51 @@
...
+	emake test
+}
+
+multilib_src_install_all() {
+	einstalldocs
+	prune_libtool_files
+
+	if use examples; then
+		find examples/ -name "Makefile.*" -delete || die
+		dodoc -r examples/
```
mgorny: Trailing slash triggers UB with new POSIX. Either use `examples` or `examples/.`, depending on the desired result.

dev-libs/libconfig/files/libconfig-1.5-out-of-source-build.patch
jbergstroem, line 1: On my todo.

mgorny:
```
+  26 May 2015; Michał Górny <mgorny@gentoo.org> libconfig-1.5.ebuild:
+  Simplify example install code, from https://github.com/gentoo/gentoo-portage-
+  rsync-mirror/pull/123 by jbergstroem.
```

jer seems to done the bump, so just merged your example improvements in.

PR #126, opened 2015-05-24
sys-freebsd/*9.2* clean up and freebsd-lib-10.1 changed to UPSTREAM_PATCHES.

nigoro:
9.2 is old, no longer used.
Please remove {9.2,9.2_rc1,9.2_rc2,9.2_rc3}.

In addition, I've deleted freebsd-lib-10.1-cve-2014-8611.patch from freebsd-lib-10.1.
I've defined UPSTREAM_PATCHES instead.

Note, about other old version.
Probably, 8.2 is required for sparc-fbsd.
9.1 is required for amd64-fbsd new user. Latest stages is 9.1...


mgorny:
Done, thanks. By the way, any reason to keep 9.1 if 9.2 is removed?

nigoro:
@mgorny,
9.1 is required when a new user (amd64-fbsd) to install.
Refer to the wiki, install 9.1, and reboot.
And, it is safest to upgrade to 10.1.

After 10.1 of stages became officially available by someone, I plan to remove the rest of the old version.

Wiki,
https://wiki.gentoo.org/wiki/Gentoo_FreeBSD

PR #124, opened 2015-05-18
Version bump: net-libs/nodejs-0.12.4

jbergstroem:
Introduce some minor functionality, otherwise straightforward.

PR #127, opened 2015-05-24
Version bump: net-libs/iojs 2.1.0

jbergstroem:
Including two things from #124. Dropping the pkg-config patch since we have a improved version upstream.


mgorny:
```
+*iojs-2.1.0 (26 May 2015)
+
+  26 May 2015; Michał Górny <mgorny@gentoo.org> +iojs-2.1.0.ebuild:
+  Version bump. Adapt gyp-nodejs symlink structure. https://github.com/gentoo
+  /gentoo-portage-rsync-mirror/pull/127 by jbergstroem.
```

PR #128, opened 2015-05-25
linux-info.eclass: force exporting CONFIG_CHECK variable to global scope

KetothXupack:
Hi all,

While writing a script which do some sanity checks for installed packages with kernel flag dependencies I found that some ebuilds define `CONFIG_CHECK` in `local` scope. This makes analysis of `/var/db/pkg/*/environment.bz2` more tricky because of missing `declare CONFIG_CHECK` record.

### DIFF
--- a/eclass/linux-info.eclass
+++ b/eclass/linux-info.eclass
@@ -912,6 +912,8 @@ linux-info_pkg_setup() {
 	fi
 
 	if [ -n "${CONFIG_CHECK}" ]; then
+		# CONFIG_CHECK is re-declared here to ensure that this variable will be
+		# available in environment.bz2
 		declare -g CONFIG_CHECK="${CONFIG_CHECK}"
 		check_extra_config
 	fi
--- a/eclass/linux-info.eclass
+++ b/eclass/linux-info.eclass
@@ -680,6 +680,10 @@ check_modules_supported() {
 # It checks the kernel config options specified by CONFIG_CHECK. It dies only when a required config option (i.e.
 # the prefix ~ is not used) doesn't satisfy the directive.
 check_extra_config() {
+	# CONFIG_CHECK is re-declared here to ensure that this variable will be
+	# available in environment.bz2
+	declare -g CONFIG_CHECK="${CONFIG_CHECK}"
+
 	local config negate die error reworkmodulenames
 	local soft_errors_count=0 hard_errors_count=0 config_required=0
 	# store the value of the QA check, because otherwise we won't catch usages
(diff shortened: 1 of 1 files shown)


mgorny:
Are you sure it really works like that? It looks to me like it just redeclares the same var for the call scope.

KetothXupack:
Hmm... 

You totally right! I've tested on wrong package. Actually `declare -g` does the trick. Unless somebody will declare `CONFIG_CHECK` as read-only :/

KetothXupack:
There is another way - declaring brand new variable. But I think that polluting global scope is not a good idea.

KetothXupack:
After revisiting usages of `linux-info.eclass` I finally realized that `check_extra_config` sometimes called directly.

KetothXupack:
Finally even this solution does not works in all cases.
For example `=media-sound/pulseaudio-5.0-r7` using `check_extra_config` function in `pkg_pretend` and somehow assignment to `CONFIG_CHECK` variable lost (maybe `pretend` executes in subshell?) and does not appears is `environment.bz2`. To be honest I'm not sure if this is a bug or declared behavior. 
I've also tried a couple of another approaches but sadly enough, seems there is no way to solve my problem at the ebuild level :disappointed:.

KetothXupack:
Finally found in [PMS#11.2](https://dev.gentoo.org/~ulm/pms/head/pms.html#x1-12200011.2)  that `pkg_pretend` really does not take part in environment saving.

So using `check_extra_config` on `pretend` stage will cause loosing `CONFIG_CHECK` value. And this is kinda bad news... The only way to validate that kernel flags are properly configured is to run `ebuild <path/to/package.ebuild> setup clean`.

mgorny:
Well, a few suggestions:
1. I don't think you can really portably turn a local variable back into a global variable. If you really want to store the values somewhere in the environment, I suggest copying them to a dedicated variable. It doesn't need to be exported, just global.
2. I think some ebuilds may call the `check_extra_config` function more than once. I guess you will want to concatenate the results in your variable. I suggest using a bash array to do that efficiently.
3. Lastly, since `pkg_pretend` is done early, ebuilds are rather expected to repeat their checks in `pkg_setup` to make sure they are still valid. If they don't do that, I think it would be reasonable to file a bug and/or a pull request. But I'll check that for sure.

KetothXupack:
Yep I tried to export new variable and accumulate CONFIG_CHECK values. But as I said it not works in all cases. For example `=media-sound/pulseaudio/pulseaudio-5.0-r4` does different checks  on `pkg_pretend` and `pkg_setup`.

Since `pretend` does not exports environment variables there is no way to do universal solution. Fixing inconsistent ebuilds is a good idea but there is always a chance that things will go wrong :)

Sadly enough, but I came across to another solution by using some magic around  `ebuild` tool instead of inspection of `environment.bz2` content. 

So... I think this pull request should be rejected and closed.

mgorny:
As you wish.

You may also replace the eclass locally, and make it output the results to some well-known location. Though you may need to use `addwrite` for sandbox.

PR #135, opened 2015-05-31
sci-misc/boinc: Version bump, fixes Bug #492602 + Fix slotting

marbre:
In reaction to https://bugs.gentoo.org/show_bug.cgi?id=492602, I took a look into the ebuilds attached to the bug report. The bump is mainly based on the attached ebuilds.
Instead of depending on `x11-libs/wxGTK:3.0[X,opengl,webkit]`, I rolled back to `x11-libs/wxGTK:2.8[X,opengl]`. Otherwise the dependencies are unfulfilled on ~sparc and ~ia64. Furthermore, I added `use systemd &&` in front to the `systemd_dounit` call and fixed slotting (also for the older versions).
On my machine, I need to uncomment the `WANT_AUTOMAKE="1.11"` to get boinc compiled.

repoman moans  about `LIVEVCS.unmasked              1`, which shouldn't be a problem since we explicitly fetch an `EGIT_COMMIT`. Anyway, maybe somebody is willing to put the source into it's dev space, as @jlec did for version 7.2.0.


mrueg:
You could use https://github.com/BOINC/boinc/releases instead of requiring git eclasses here. :)

marbre:
@mrueg You're absolutely right. Didn't thought about it ;)
Fixed this and further readded the systemd useflag.

mgorny:
@jlec, want to look at it with sci eye?

sci-misc/boinc/boinc-7.2.0-r1.ebuild
```diff
@@ -23,7 +23,7 @@ RDEPEND="
 	!sci-misc/boinc-bin
 	!app-admin/quickswitch
 	>=app-misc/ca-certificates-20080809
-	dev-libs/openssl
+	dev-libs/openssl:*
```
mgorny: Does boinc use openssl at build time? If so, it needs OpenSSL headers and `:*` is invalid then since other slots don't provide the headers.

sci-misc/boinc/boinc-7.2.0-r1.ebuild
```diff
@@ -35,7 +35,7 @@ RDEPEND="
 		dev-db/sqlite:3
 		media-libs/freeglut
 		sys-libs/glibc:2.2
-		virtual/jpeg
+		virtual/jpeg:*
```
mgorny: Ditto.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,147 @@
...
+		newicon "${S}"/packages/generic/sea/${PN}mgr.48x48.png ${PN}.png || die
+		make_desktop_entry boincmgr "${PN}" "${PN}" "Math;Science" "Path=/var/lib/${PN}"
+	fi
+
+	# cleanup cruft
+	rm -rf "${ED}"/etc/
+
+	newinitd "${FILESDIR}"/${PN}.init ${PN}
+	newconfd "${FILESDIR}"/${PN}.conf ${PN}
+	use systemd && systemd_dounit "${FILESDIR}"/${PN}.service
```
mgorny: This is incorrect. systemd units are installed unconditionally.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,147 @@
...
+}
+
+pkg_preinst() {
+	enewgroup ${PN}
+	# note this works only for first install so we have to
+	# elog user about the need of being in video group
+	if use cuda; then
```
mgorny, line 114: I'd suggest rolling that into one `enewuser` call with groups set in separate variable.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,147 @@
...
+		enewuser ${PN} -1 -1 /var/lib/${PN} "${PN},video"
+	else
+		enewuser ${PN} -1 -1 /var/lib/${PN} "${PN}"
+	fi
+}
+
+pkg_postinst() {
+	echo
```
mgorny, line 121: Don't mix echo & elog.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,147 @@
...
+	elog "You are using the source compiled version of ${PN}."
```
mgorny: It'd be better to replace most of ${PN} with verbose boinc. It will make it easier to read this and to check line width.

sci-misc/boinc/boinc-7.4.42.ebuild
jlec, line 121: I would say, use "echo" for simple black lines and "elog/einfo" when the line has content.

sci-misc/boinc/boinc-7.2.0-r1.ebuild
marbre: A realy short look into the [configure.ac](https://github.com/BOINC/boinc/blob/master/configure.ac) showed that ssl and jpeg are checked at buildtime. Thus `:=` would be the right here and for jpeg?

I think I misunderstood `:*`. The [wiki](https://devmanual.gentoo.org/general-concepts/dependencies/) syas that `:* means that any slot is acceptable, and explicitly declares that changes in the slot or sub-slot can be ignored;`. I thought that means that changes within the sub-slot can be ignored, which would be okay. Anyway, also if the slot is changed the package depending on it isn't recompiled, correct? Further `:=` leads to recompiling of the package if openssl changes.
Please correct me if I am wrong again.

sci-misc/boinc/boinc-7.4.42.ebuild
marbre: Done.

sci-misc/boinc/boinc-7.4.42.ebuild
marbre, line 114: Hope, that's okay(?).

sci-misc/boinc/boinc-7.4.42.ebuild
marbre, line 121: Thus, no change required here?

sci-misc/boinc/boinc-7.4.42.ebuild
marbre: Done.

sci-misc/boinc/boinc-7.2.0-r1.ebuild
```diff
@@ -23,7 +23,7 @@ RDEPEND="
 	!sci-misc/boinc-bin
 	!app-admin/quickswitch
 	>=app-misc/ca-certificates-20080809
-	dev-libs/openssl
+	dev-libs/openssl:=
```
prometheanfire: is there a reason you did not go with `:*` instead of `:=`?

graaff: Actually this should be "dev-libs/openssl:0". The other slot only exists for binary compatibility and should not be used for normal packages.

mgorny: Actually `dev-libs/openssl:0=` to cover the next time SONAME changes.

marbre: @prometheanfire That was my first idea, but since @mgorny commented in the first line note ("Does boinc use openssl at build time? If so, it needs OpenSSL headers and :\* is invalid then since other slots don't provide the headers.") I changed it to `:=`.
Anyway, now I changed it once again to `:0=`. I still need to get deeper into slotting..

sci-misc/boinc/boinc-7.2.0-r1.ebuild
```diff
@@ -35,7 +35,7 @@ RDEPEND="
 		dev-db/sqlite:3
 		media-libs/freeglut
 		sys-libs/glibc:2.2
-		virtual/jpeg
+		virtual/jpeg:=
```
mgorny: You want `:0=` here as well. `virtual/jpeg:62` is binary-only.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,147 @@
...
+
+src_prepare() {
+	# prevent bad changes in compile flags, bug 286701
+	sed -i -e "s:BOINC_SET_COMPILE_FLAGS::" configure.ac || die "sed failed"
+
+	autotools-utils_src_prepare
+}
+
+src_configure() {
+	local wxconf=""
```
mgorny: Please use an array here so that we can support whitespace in paths properly. Not that it matters right here, but it's cleaner this way.

mgorny: In fact, you could drop the whole wxconf thing and just append to myeconfargs.

sci-misc/boinc/boinc-7.2.0-r1.ebuild
marbre: Done.

sci-misc/boinc/boinc-7.4.42.ebuild
marbre: Fixed for version 7.4.42.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+# Distributed under the terms of the GNU General Public License v2
+# $Header: $
+
+EAPI=5
+
+#WANT_AUTOMAKE="1.11"
+
+AUTOTOOLS_AUTORECONF=true
+
+inherit autotools-utils eutils flag-o-matic systemd user versionator wxwidgets
```
mgorny, line 11: `eutils` seems unused here.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+MY_PV=$(get_version_component_range 1-2)
+
+DESCRIPTION="The Berkeley Open Infrastructure for Network Computing"
+HOMEPAGE="http://boinc.ssl.berkeley.edu/"
+SRC_URI="https://github.com/BOINC/boinc/archive/client_release/${MY_PV}/${PV}.tar.gz -> ${P}.tar.gz"
+
+LICENSE="LGPL-2.1"
+SLOT="0"
+KEYWORDS="~amd64 ~ia64 ~ppc ~ppc64 ~sparc ~x86"
+IUSE="X cuda static-libs systemd"
```
mgorny, line 22: `USE=systemd` is unused.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+		--with-ssl
+		$(use_with X x)
+		$(use_enable X manager)
+	)
+
+	# look for wxGTK
+	if use X; then
+		WX_GTK_VER="2.8"
+		need-wxwidgets unicode
+		myeconfargs+=(--with-wx-config=${WX_CONFIG})
```
mgorny, line 82: Missing quoting around path in `WX_CONFIG`.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+		myeconfargs+=(--without-wxdir)
+	fi
+
+	autotools-utils_src_configure
+}
+
+src_install() {
+	autotools-utils_src_install
+
+	dodir /var/lib/${PN}/
```
mgorny, line 93: `dodir` is redundant, `keepdir` creates the directory.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+	keepdir /var/lib/${PN}/
```
mgorny, line 94: And the trailing slash is a bad idea here.

sci-misc/boinc/boinc-7.4.42.ebuild
mgorny, line 11: Sorry, I was wrong. I see `newicon` and `make_desktop_entry` now.

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+
+	if use X; then
+		newicon "${S}"/packages/generic/sea/${PN}mgr.48x48.png ${PN}.png || die
```
mgorny, line 97: `die` is unnecessary here (`newicon` dies on its own)

sci-misc/boinc/boinc-7.4.42.ebuild
```diff
@@ -0,0 +1,145 @@
...
+src_prepare() {
+	# prevent bad changes in compile flags, bug 286701
+	sed -i -e "s:BOINC_SET_COMPILE_FLAGS::" configure.ac || die "sed failed"
+
+	autotools-utils_src_prepare
+}
+
+src_configure() {
+	# add gtk includes
+	use X && append-flags "$(pkg-config --cflags gtk+-2.0)"
```
mgorny, line 65: This may be unnecessary judging by `Makefile.am`. Though they have a typo in there but I'll fix it.

mgorny:
That all was FYI, I'm going to fix those minor mistakes and commit it finally.

mgorny:
```
+*boinc-7.4.42 (04 Aug 2015)
+
+  04 Aug 2015; Michał Górny <mgorny@gentoo.org> +boinc-7.4.42.ebuild,
+  boinc-7.2.0-r1.ebuild, boinc-7.2.0.ebuild:
+  Version bump with minor fixes. https://github.com/gentoo/gentoo-portage-rsync-
+  mirror/pull/135 by marbre.
```

PR #136, opened 2015-06-01
version bump: dev-libs/hiredis 0.13.1

jbergstroem:
Bump including a few fixes and features. 

question: does insinto/doins require ${ED}?


jbergstroem:
Not sure about returning 0/1 vs die in src_test. Would appreciate some feedback on improving that as well.

jbergstroem:
Also, how do we handle the dep to dev-db/redis for the other arches?

jbergstroem:
bringing @neurogeek (maintainer) in here.

jbergstroem:
ping @mgorny

dev-libs/hiredis/files/hiredis-0.13.1-disable-network-tests.patch
```diff
@@ -0,0 +1,36 @@
+diff --git test.c test.c
+index 8fde554..89ed6a0 100644
+--- test.c
++++ test.c
+@@ -343,6 +343,7 @@ static void test_free_null(void) {
+ static void test_blocking_connection_errors(void) {
+     redisContext *c;
+ 
++    /*
```
mgorny: Safer to do `#if 0`. Won't break on comments in the middle of it :).

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,73 @@
...
+DEPEND="test? ( dev-db/redis )"
+
+src_prepare() {
+	epatch "${FILESDIR}/${P}-disable-network-tests.patch"
+
+	# use GNU ld syntax on Solaris
+	sed -i -e '/DYLIB_MAKE_CMD=.* -G/d' Makefile || die
+}
+
+_emake() {
```
mgorny: Please use a more verbose name. This one's easy to confuse.

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,73 @@
...
+	local REDIS_PORT=56379
+	local REDIS_TEST_CONFIG="daemonize yes \n
+		pidfile ${REDIS_PID} \n
+		port ${REDIS_PORT} \n
+		bind 127.0.0.1 \n
+		unixsocket //${REDIS_SOCK}"
+
+	_emake hiredis-test
+
+	echo -e ${REDIS_TEST_CONFIG} | /usr/sbin/redis-server - || die
```
mgorny: No, no, no, no, no, no, no, no, no.

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,73 @@
...
+	# overhead is low w/creating it all the time.  It's also needed
+	# by the tests.
+	_emake dynamic static hiredis.pc
+}
+
+src_test() {
+	local REDIS_PID="${T}"/hiredis.pid
+	local REDIS_SOCK="${T}"/hiredis.sock
+	local REDIS_PORT=56379
+	local REDIS_TEST_CONFIG="daemonize yes \n
```
mgorny: Use bash's `$""` syntax. Or just be happy with the inline newlines you're doing anyway.

dev-libs/hiredis/hiredis-0.13.1.ebuild
mgorny: So no `echo -e`, and I suggest using some variation of `<` (`<<<`?) to have redis-server as the top process rather than the subprocess.

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,73 @@
...
+		unixsocket //${REDIS_SOCK}"
+
+	_emake hiredis-test
+
+	echo -e ${REDIS_TEST_CONFIG} | /usr/sbin/redis-server - || die
+	./hiredis-test -h 127.0.0.1 -p ${REDIS_PID} -s ${REDIS_SOCK} || die
+	local ret=$?
+
+	kill "$(<"${REDIS_PID}")" || die
+	return ${ret}
```
mgorny: This is 100% meaningless. Phase function returns are ignored.

jbergstroem: that was one of my questions above.

jbergstroem:
So, how do we handle KEYWORDS between hiredis and redis?

mgorny:
Well, obviously you drop the keywords that are not on redis.

mgorny:
Now, if you really believe someone used hiredis on those arches, we need a new keywordreq for both packages.

jbergstroem:
@mgorny ..or expand dev-db/redis? I can't vouch wider than x86/amd64. Keywords seems to been around for a while (besides alpha) so it'd be great to get @neurogeek's opinion.

jbergstroem:
Just added an attempt to fix the paths in the pkg-config file. I feel like it isn't quite enough though;
- how would I handle prefix installs?
- the generated .pc seems to hardcode lib path, should i rewrite that to use libdir instead?
  
  ``` bash
  $  pkg-config --libs hiredis                                             
  -L/usr/lib -lhiredis 
  ```

mgorny:
I guess you'd pass `${EPREFIX}/usr` on Prefix but I'm not sure how exactly `PREFIX` works in the package. And yep, you gotta fix the libdir in pkg-config file.

mgorny:
Just to be clear, if you're not sure about Prefix support, you can leave it out. Prefix is pretty much 'if we can, we do it, if we can't, we leave it to prefix@ to fix :)'.

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	# overhead is low w/creating it all the time.  It's also needed
+	# by the tests.
+	_build dynamic static hiredis.pc
+}
+
+src_test() {
+	local REDIS_PID="${T}"/hiredis.pid
+	local REDIS_SOCK="${T}"/hiredis.sock
+	local REDIS_PORT=56379
+	local REDIS_TEST_CONFIG=$"daemonize yes
```
mgorny: Well, `$""` implies parsing escapes like \n. If you don't use them anymore, use plain `""` :).

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	./hiredis-test -h 127.0.0.1 -p ${REDIS_PID} -s ${REDIS_SOCK}
+	local ret=$?
+
+	kill "$(<"${REDIS_PID}")" || die
+	[ ${ret} != "0" ] && die "tests failed"
+}
+
+src_install() {
+	_build PREFIX="${ED%/}/usr" install
+	use static-libs || rm "${ED%/}/usr/$(get_libdir)/libhiredis.a"
```
mgorny: rm||die :).

dev-libs/hiredis/hiredis-0.13.1.ebuild
```diff
@@ -0,0 +1,75 @@
...
+	use examples && dohtml -r examples
```
mgorny: `dohtml` is deprecated, and looks weird here anyway. Use `dodoc`.

dev-libs/hiredis/hiredis-0.13.1.ebuild
jbergstroem: The config assumes each option is on a newline

mgorny: Yes, but you don't need `$""` for the inline newlines.

mgorny:
@jbergstroem, are you still there? Also you need to rebase :P.

jbergstroem:
Rebased and fixed outstanding issues.

mgorny:
```
+*hiredis-0.13.1 (15 Jun 2015)
+
+  15 Jun 2015; Michał Górny <mgorny@gentoo.org>
+  +files/hiredis-0.13.1-disable-network-tests.patch, +hiredis-0.13.1.ebuild:
+  Version bump with some fixes. Install a pkg-config file. Run tests against a
+  isolated redis instance.
```

Also:

```
  KEYWORDS.dropped              1
   dev-libs/hiredis/hiredis-0.13.1.ebuild: alpha arm ia64 ppc sparc
```

Now you go file a keywordreq :).

PR #139, opened 2015-06-02
Fixed bug 545714, tar in some cases fail to decompress of tar.xz file on Gentoo/FreeBSD.

nigoro:
https://bugs.gentoo.org/show_bug.cgi?id=545714

Fix the following problems.

```
# tar -xJpf /var/tmp/distfiles/freebsd-src-10.1.tar.xz -C /var/tmp/portage/ --strip-components=2 usr/src/share
tar: Truncated input file (needed 1024 bytes, only 0 available)
tar: Error exit delayed from previous errors.

# tar -xJpf /var/tmp/distfiles/freebsd-src-10.1.tar.xz -C /var/tmp/portage/
usr/src/tools/tools/mid/mid-build: Lzma library error: Corrupted input data
tar: Error exit delayed from previous errors.
```


app-arch/xz-utils/xz-utils-5.2.1-r1.ebuild
```diff
@@ -28,7 +28,7 @@ HOMEPAGE="http://tukaani.org/xz/"
 # See top-level COPYING file as it outlines the various pieces and their licenses.
 LICENSE="public-domain LGPL-2.1+ GPL-2+"
 SLOT="0"
-IUSE="nls static-libs +threads"
+IUSE="kernel_FreeBSD nls static-libs +threads"
```
mgorny: Wouldn't userland_\* be more correct here? Assuming the `sha256.h` header is userland, not kernel stuff :).

nigoro:
@mgorny 
thanks for your comment.
userland_BSD to affect all of BSD. 
I don't know need this fixed in the other BSD...

Changed from kernel_FreeBSD to elibc_FreeBSD.
Because sha256.h has been provided by -lib.

```
# equery b sha256.h
 * Searching for sha256.h ...
sys-freebsd/freebsd-lib-10.1 (/usr/include/sha256.h)
```

mgorny:
Thanks, this seems even more correct.

mgorny:
```
+*xz-utils-5.2.1-r1 (03 Jun 2015)
+
+  03 Jun 2015; Michał Górny <mgorny@gentoo.org> +xz-utils-5.2.1-r1.ebuild,
+  -xz-utils-5.2.1.ebuild, xz-utils-9999.ebuild:
+  Disable sha256.h on *BSD, bug #545714. https://github.com/gentoo/gentoo-
+  portage-rsync-mirror/pull/139 by nigoro.
```

PR #140, opened 2015-06-03
llvm-3.6.1[clang]: fix build fails with using gcc-4.9.2 on GFBSD

nigoro:
https://bugs.gentoo.org/show_bug.cgi?id=548444

I want add KEYWORDS="~amd64-fbsd ~x86-fbsd" to gcc-4.9.2.
However, llvm-3.6.x[clang] fails to compile when using gcc-4.9.2.
This PR will fix the issue.

Note,
Fixed in trunk on upstream. 
This patch in llvm-3.7 will become unnecessary.

See also,
https://llvm.org/bugs/show_bug.cgi?id=22302
http://llvm.org/viewvc/llvm-project/cfe/trunk/lib/Driver/MSVCToolChain.cpp?r1=223391&r2=226925


sys-devel/llvm/llvm-3.6.1.ebuild
```diff
@@ -160,6 +160,9 @@ src_prepare() {
 		eprefixify tools/clang/lib/Frontend/InitHeaderSearch.cpp
 	fi
 
+	# Fix build fails with using gcc-4.9 on Gentoo/FreeBSD, bug #548444
+	use clang && epatch "${FILESDIR}"/clang-3.6-fbsd-gcc49.patch
```
mgorny: We have a nice `use clang` block a few lines up :).

nigoro:
move epatch clang-3.6-fbsd-gcc49.patch ;-)

mgorny:
```
+  14 Jun 2015; Michał Górny <mgorny@gentoo.org>
+  +files/clang-3.6-fbsd-gcc49.patch, llvm-3.6.1.ebuild:
+  Apply upstream patch for FreeBSD gcc-4.9 build, bug #548444.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/140 by nigoro.
```

PR #142, opened 2015-06-04
x11-libs/wxGTK: multilib support

nE0sIghT:
Multilib support for wxGTK 2.8 and 3.0 as i requested here: https://bugs.gentoo.org/show_bug.cgi?id=510708
This changes needed for games-emulation/pcsx2: https://bugs.gentoo.org/show_bug.cgi?id=510710

I added multilib support to eselect-wxwidgets via epatch here. This change available also at https://github.com/nE0sIghT/eselect/commits/eselect-wxwidgets-multilib and must be applied to eselect main repository

### DIFF
--- a/x11-libs/wxGTK/files/wxGTK-2.8.12.1-bakefile.patch
+++ b/x11-libs/wxGTK/files/wxGTK-2.8.12.1-bakefile.patch
@@ -1,4 +1,4 @@
-From baa08e14da5fe588bc0e4fa53cd6471a328c6e11 Mon Sep 17 00:00:00 2001
+From 006b0650b0f390c1e9a19361988b6af7a327e296 Mon Sep 17 00:00:00 2001
 From: nE0sIghT <ykonotopov@gmail.com>
 Date: Wed, 5 Aug 2015 12:38:39 +0300
 Subject: [PATCH 2/2] Fixed bakefile version
--- a/x11-libs/wxGTK/files/wxGTK-2.8.12.1-bakefile.patch
+++ b/x11-libs/wxGTK/files/wxGTK-2.8.12.1-bakefile.patch
@@ -1,0 +1,25 @@
+From baa08e14da5fe588bc0e4fa53cd6471a328c6e11 Mon Sep 17 00:00:00 2001
+From: nE0sIghT <ykonotopov@gmail.com>
+Date: Wed, 5 Aug 2015 12:38:39 +0300
+Subject: [PATCH 2/2] Fixed bakefile version
+
+---
+ build/aclocal/bakefile.m4 | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/build/aclocal/bakefile.m4 b/build/aclocal/bakefile.m4
+index 88478ca..aa04b16 100644
+--- a/build/aclocal/bakefile.m4
++++ b/build/aclocal/bakefile.m4
+@@ -837,7 +837,7 @@ AC_DEFUN([AC_BAKEFILE],
+     AC_SUBST(OBJCXXFLAGS)
+ 
+ 
+-    BAKEFILE_BAKEFILE_M4_VERSION="0.2.8"
++    BAKEFILE_BAKEFILE_M4_VERSION="0.2.9"
+ 
+     dnl includes autoconf_inc.m4:
+     $1
+-- 
+2.4.6
+
--- a/x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
+++ b/x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
@@ -68,6 +68,10 @@ LICENSE="wxWinLL-3
 
 S="${WORKDIR}/wxPython-src-${PV}"
 
+MULTILIB_CHOST_TOOLS=(
+	usr/bin/wx-config-2.8
+)
+
 src_prepare() {
 	epatch "${FILESDIR}"/${PN}-2.8.11-unicode-odbc.patch
 	epatch "${FILESDIR}"/${PN}-2.8.11-collision.patch
@@ -160,6 +164,13 @@ multilib_src_install() {
 		cd contrib/src || die
 		emake DESTDIR="${D}" install
 	fi
+
+	# Manually symlink wxconfig tool to make MULTILIB_CHOST_TOOLS work
+	local wxconfig=$(readlink "${ED}${MULTILIB_CHOST_TOOLS[0]}")
+	local wxconfig_dir=${wxconfig%/*}
+	local wxconfig_name=${wxconfig##*/}
+
+	ln -s "${wxconfig_name}" "${ED}${wxconfig_dir}/${CHOST}-${wxconfig_name}" || die
 }
 
 multilib_src_install_all() {
--- a/x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
+++ b/x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
@@ -3,8 +3,9 @@
 # $Header: /var/cvsroot/gentoo-x86/x11-libs/wxGTK/wxGTK-2.8.12.1-r1.ebuild,v 1.13 2015/03/31 20:05:20 ulm Exp $
 
 EAPI="5"
+WANT_AUTOCONF=2.5
 
-inherit eutils flag-o-matic versionator multilib-minimal
+inherit autotools eutils flag-o-matic versionator multilib-minimal
 
 DESCRIPTION="GTK+ version of wxWidgets, a cross-platform C++ GUI toolkit"
 HOMEPAGE="http://wxwidgets.org/"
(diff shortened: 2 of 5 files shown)


mgorny:
@axs-gentoo, could you take a look at this one, please? :)

mgorny:
@leio!

app-eselect/eselect-wxwidgets/files/0001-wxwidgets.eselect-multilib-support.patch
```diff
@@ -0,0 +1,277 @@
...
+ 
+ DESCRIPTION="Manage the system default wxWidgets profile"
+ MAINTAINER="wxwidgets@gentoo.org"
+-VERSION="20140423"
++VERSION="20150329"
+ 
+ WXCONFFILE="${EROOT}"/var/lib/wxwidgets/current
+-WXCONFDIR="${EROOT}"/usr/lib/wx/config
++WXCONFDIR_64="${EROOT}"/usr/lib/wx/config
++WXCONFDIR_32="${EROOT}"/usr/lib32/wx/config
```
mgorny: No, no, no. There's some eselect library for multilib support. If you're doing that, you need to support all ABIs, not just x86+amd64.

x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
```diff
@@ -0,0 +1,187 @@
...
+	fi
+
+	econf ${myconf}
+}
+
+multilib_src_compile() {
+	emake
+
+	if [[ -d contrib/src ]]; then
+		cd contrib/src
```
mgorny: ||die

x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
```diff
@@ -0,0 +1,187 @@
...
+	default
+
+	if [[ -d contrib/src ]]; then
+		cd contrib/src
+		emake DESTDIR="${D}" install
+	fi
+}
+
+multilib_src_install_all() {
+	cd "${S}"/docs
```
mgorny: ||die :)

x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
```diff
@@ -0,0 +1,187 @@
...
+	dodoc changes.txt readme.txt todo30.txt
+	newdoc base/readme.txt base_readme.txt
+	newdoc gtk/readme.txt gtk_readme.txt
+
+	if use doc; then
+		dohtml -r "${S}"/docs/html/*
```
mgorny: dohtml is deprecated, use `dodoc -r "${S}"/docs/html`

nE0sIghT:
Pushed fixes.
Branch eselect-wxwidgets-multilib at https://github.com/nE0sIghT/eselect overwritten

nE0sIghT:
> And anyway, I'm not sure if it's a good idea to allow combining different wxwidgets versions for different ABIs. The module would certainly be simpler if you just set the same thing for all libdirs (this is e.g. how eselect-opengl does it).

So maybe drop eselect-wxwidgets multilib changes?
It can be used only for out-of-portage builds (and maybe for broken portage builds, but i don't think there is exists such) that relies on wxconfig utility in $PATH.

mgorny:
That's some option too. Especially that the current eselect module is broken for me anyway.

axs-gentoo:
Still need anything from me here?  It seems all issues have been resolved now.  I second the dropping of multilib changes to eselect-wxwidgets.

mgorny:
Yes, I only looked at the eselect module :P. I would really appreciate if you could take a look at this one, so I could focus on the repo-mirror-ci scripting.

mgorny:
Sorry for replying this late. I don't really like the way get_libdir is hacked in there. Could you by any chance look into replacing the sed with something better? Possibly an upstreamable patch?

nE0sIghT:
I replaced sed fix with configure patch in last commit.
Patch for wxGTK-2.8 is huge because of whitespace/quotes changes in regenerated configure script... Do not know how to avoid this (i used same version of autoconf - 2.59)

nE0sIghT:
Upstream PR: https://github.com/wxWidgets/wxWidgets/pull/59

nE0sIghT:
[Upstream said](https://github.com/wxWidgets/wxWidgets/pull/59#issuecomment-127772039) that libdir problems should not affect 3.0 wxGTK.
I tested x32 chroot and it compiled well (and linked - as i see) without any sed/patch. However i can not [test ia64](https://bugs.gentoo.org/show_bug.cgi?id=499984), but that bug was reported against 2.8 version also.

mgorny:
Well, I personally think all that magic should be killed with fire. If possible, please try to backport the upstream change to 2.8. That would certainly be much cleaner. After that, it should be possible to build wxGTK out-of-source.

nE0sIghT:
I [tried to backport](https://github.com/nE0sIghT/wxWidgets/commits/feature/multilib) library search code and was able to pass configure stage, but now i hit strange relocation/-fPIC problem...

nE0sIghT:
Problem was in autoconf.eclass - it uses autoconf-2.69 when WANT_AUTOCONF=2.5 specified.
I pushed commit with backported patch - is it ok to depend and directly call autoconf-2.59?

mgorny:
Nope, that's not allowed. I think you need to figure out how to make it happy about new autoconf.

nE0sIghT:
I added autoconf patch and rebased PR.
Tested with ABI_X86="32 64 x32"

x11-libs/wxGTK/files/wxGTK-2.8.12.1-bakefile.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/build/aclocal/bakefile.m4 b/build/aclocal/bakefile.m4
+index 88478ca..aa04b16 100644
+--- a/build/aclocal/bakefile.m4
++++ b/build/aclocal/bakefile.m4
+@@ -837,7 +837,7 @@ AC_DEFUN([AC_BAKEFILE],
+     AC_SUBST(OBJCXXFLAGS)
+ 
+ 
+-    BAKEFILE_BAKEFILE_M4_VERSION="0.2.8"
++    BAKEFILE_BAKEFILE_M4_VERSION="0.2.9"
```
mgorny, line 19: Just to be clear, does this really matter?

mgorny:
Great work! I'm going to test it now, and hopefully commit it soon enough :).

Do you want to become wxGTK co-maintainer by any chance? You seem to be handling this creepy thing great, and wxwidgets team would certainly use some help.

x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
```diff
@@ -0,0 +1,189 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Header: /var/cvsroot/gentoo-x86/x11-libs/wxGTK/wxGTK-2.8.12.1-r1.ebuild,v 1.13 2015/03/31 20:05:20 ulm Exp $
+
+EAPI="5"
+WANT_AUTOCONF=2.5
```
mgorny: Do you actually need this anymore?

mgorny:
Bad news (2.8), 32&64:

```
checking for Xinerama... no
configure: WARNING: Xinerama not found; disabling wxDisplay
checking for -lSM - X11 session management... no
configure: WARNING: libSM not found; disabling session management detection
```

I think you need to backport a bit more :).

x11-libs/wxGTK/files/wxGTK-2.8.12.1-bakefile.patch
nE0sIghT, line 19: Yes. Without that change build failed with

```
configure: error: Versions of Bakefile used to generate makefiles (0.2.9) and configure (0.2.8) do not match.
```

x11-libs/wxGTK/wxGTK-2.8.12.1-r2.ebuild
nE0sIghT: I added this because with next autoconf release build may break again

mgorny: It will be a build-time breakage, so better than forcing dep that people will forget to update :).

mgorny:
You also want to `MULTILIB_CHOST_TOOLS` wx-config-3.0.

nE0sIghT:
I updated backported patch for Xinerama and SM and removed autoconf dependency.
Also added MULTILIB_CHOST_TOOLS part, but it has turned a bit tricky because wx-config tools already laid in libdir.

> Do you want to become wxGTK co-maintainer by any chance?

I will be glad to help if i can :-)

mgorny:
Ouch. Let's revert that chost tool part then, I see that the eclass uses /usr/lib\* directly anyway. I'll just revert it myself while committing. Now I'm going to do some live surgery on my partitions, then I'll get back to you and test 2.8 again.

mgorny:
```
+*wxGTK-3.0.2.0-r2 (06 Aug 2015)
+*wxGTK-2.8.12.1-r2 (06 Aug 2015)
+
+  06 Aug 2015; Michał Górny <mgorny@gentoo.org>
+  +files/wxGTK-2.8.12.1-autoconf.patch, +files/wxGTK-2.8.12.1-bakefile.patch,
+  +files/wxGTK-2.8.12.1-libdir.patch, +wxGTK-2.8.12.1-r2.ebuild,
+  +wxGTK-3.0.2.0-r2.ebuild:
+  Enable multilib support. Backport some build system patches.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/142 by nE0sIghT.
```

I've also pinged wxGTK maintainers about you. I'll let you know when I get a reply.

PR #143, opened 2015-06-06
nfs.initd - fix typo (useless empty space)

mm1ke:
This is a tiny fix for nfs.initd. It kills some empty space in the file. I also haven't created an bug for this (bug i can still create one if needed) :)


mgorny:
I think it was actually intended in order to align those variables on `=`.

mgorny:
(not that I like the idea)

mm1ke:
Oh, ok, strange align - i didn't even see it until you mentioned it ;)
However, in that case you can ignore my PR.

PR #147, opened 2015-06-14
boot0-10.1: Drop DEPEND=clang. Changed to not force CC=clang.

nigoro:
Currently, it can be compiled using gcc.
I've tested gcc-4.6.4, gcc-4,8,4, gcc-4.9.2.


mgorny:
```
+  14 Jun 2015; Michał Górny <mgorny@gentoo.org> boot0-10.1.ebuild:
+  Do not force clang, gcc works nowadays. https://github.com/gentoo/gentoo-
+  portage-rsync-mirror/pull/147 by nigoro.
```

PR #149, opened 2015-06-14
freebsd-lib-{9.1-r11,10.1}: QA fix, bug 500990.

nigoro:
Fix QA Notice: The following shared libraries lack NEEDED entries /usr/lib32/libc.so.7 

https://bugs.gentoo.org/show_bug.cgi?id=500990


mgorny:
```
+  14 Jun 2015; Michał Górny <mgorny@gentoo.org> freebsd-lib-10.1.ebuild,
+  freebsd-lib-9.1-r11.ebuild:
+  Silence DT_NEEDED QA warnings, bug #500990. https://github.com/gentoo/gentoo-
+  portage-rsync-mirror/pull/149 by nigoro.
```

PR #141, opened 2015-06-04
version bump: dev-libs/libuv-1.6.0

jbergstroem:
Copy from libuv 1.5.0.

PR #152, opened 2015-06-15
Version bump: net-libs/iojs-2.3.0

jbergstroem:
Bumped dependencies, otherwise no changes. Depends on #141.


mgorny:
So when are you removing the bundled-libs crap? :P

mgorny:
```
+*iojs-2.3.0 (15 Jun 2015)
+
+  15 Jun 2015; Michał Górny <mgorny@gentoo.org> +iojs-2.3.0.ebuild:
+  Version bump.
```

jbergstroem:
I'll look at removing that once we converge node and io.js. I still fight to make share libraries a supported release upstream.

PR #153, opened 2015-06-23
version bump: net-libs/iojs-2.3.1

jbergstroem:
Added epatch_user, otherwise no changes. This [upstream version] fixes a bug some users have experienced: https://bugs.gentoo.org/show_bug.cgi?id=552232.


mgorny:
```
+*iojs-2.3.1 (25 Jun 2015)
+
+  25 Jun 2015; Michał Górny <mgorny@gentoo.org> +iojs-2.3.1.ebuild:
+  Version bump with upstream fix for bug #552232. https://github.com/gentoo
+  /gentoo-portage-rsync-mirror/pull/153 by jbergstroem.
```

Now please prepare a PR cleaning up some of the old versions :).

PR #154, opened 2015-06-24
version bump: www-servers/nginx-1.8.0

jbergstroem:
Here's the new nginx stable series. It's as close to nginx 1.7.6 as possible, only updating modules that needs compile fixes for the late api changes in the nginx 1.7.x series.

Parallel to this, I will introduce nginx 1.9.0 which will consistently update third party modules.

The most notable (and unfortunate) change is that the ajp module has compile issues that upstream hasn't fixed. I chose to delay introducing nginx 1.8.0 to gentoo for this reason, but it's been long enough. Because of this, I bail if people still have the use-flag set to make sure they don't update nginx and end up with a broken server.


mrueg:
Merged, thanks!

PR #156, opened 2015-06-26
version bump: www-servers/nginx-1.9.2

jbergstroem:
A few changes from previous version. Most noteworthy:
- now uses SLOT for the unstable series (called `mainline` upstream) -- this implies that we update other nginx ebuilds with an addition to RDEPEND
- update all packages
- introduce USE=threads
- bail early on a known upstream module bug in mogilefs (https://github.com/vkholodkov/nginx-mogilefs-module/issues/7)
- adds the more powerful (compared to the built in) [memc module](https://github.com/openresty/memc-nginx-module)


mgorny:
What's the deal with the slot? Is it co-installable with stable nginx version?

jbergstroem:
No, the idea was to allow people to install it if they know what they're doing without resorting to keywords. The mainline version tend to break which made our users unhappy during the 1.7.x cycle which I'm trying to avoid.

mgorny:
Then it's invalid use of slots.

jbergstroem:
@mrueg: thoughts on above?

mrueg:
@mgorny This seems to be a common pattern, that is used in the tree. See for example: www-plugins/chrome-binary-plugins, dev-util/jenkins-bin. x11-drivers/ati-drivers had once a legacy slot, that blocked slot :0 and vice versa. 
Although this is not the problem slotting was originally intended to solve, I don't see any issue here.

mgorny:
```
  IUSE.invalid [fatal]          1
   www-servers/nginx/nginx-1.9.2.ebuild: nginx_modules_http_memc
```

Please include the *.desc file update as well.

jbergstroem:
ok, should be good to go. I'm still not sure what to do with SLOT:ing, but since I'm only a proxy I'm happy to leave that decision to the reviewers :)

mgorny:
Why are you removing a patch that's used by 3 other versions? I'm going to fix it this time but you are supposed not to make such a stupid mistakes :p.

mgorny:
And kill that space-indentation in your editor.

mgorny:
```
+*nginx-1.9.2 (13 Jul 2015)
+
+  13 Jul 2015; Michał Górny <mgorny@gentoo.org> +files/check-1.9.2.patch,
+  +nginx-1.9.2.ebuild:
+  Version bump for mainline series. Introduce USE=threads. Add memc module.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/156 by jbergstroem.
```

jbergstroem:
The reason that patch was dropped was because this ebuild was based on the [nginx overlay](https://github.com/gentoo/nginx-overlay). I'll rebase that repo onto this forward.

PR #157, opened 2015-06-28
[media-libs/oyranos] Fix 9999

a17r:
Fixed live ebuild wanted for testing of kolor-manager


mgorny:
I think you should use `REQUIRED_USE='?? ( qt4 qt5 )'` rather than the implicit fallback. @pesa, @johu, could you confirm?

a17r:
I did it according to https://wiki.gentoo.org/wiki/Project:Qt/Policies#Optional_support_for_one_of_two_Qt_versions

mgorny:
Ok then. Sadly, this is crappy for binary packages but I guess we'd have to live with crappy policies.

Pesa:
@mgorny what exactly is crappy for binary packages?

mgorny:
The fact that `USE=-qt4 qt5` and `USE=qt4 qt5` is the same thing, albeit binary package is incompatible.

Pesa:
I wouldn't call that "crappy", it's just suboptimal ;) and in any case much much better than a portage error out-of-the-box because both qt4 and qt5 are enabled (very common scenario), resulting in many unhappy users that don't know how to handle the situation. @kensington should know more about this.

a17r:
Last bits of buildsystem patch upstreamed!

I would also recommend cleaning up:
media-libs/oyranos-0.9.4-r2 (missing RDEPEND, load of patches)
media-gfx/icc_examin-0.54 (only package that depends on 0.9.4)

kensington:
@mgorny it's definitely not ideal, but we were getting an excessive number of complaints about hitting the REQUIRED_USE and we don't really have any other options (if we keep the choice available).

media-libs/oyranos/oyranos-9999.ebuild
```diff
@@ -80,18 +87,37 @@ multilib_src_configure() {
...
+	fi
 
-		# only used in programs
-		$(multilib_native_usex fltk -DWANT_FLTK=1 "")
-		$(multilib_native_usex qt4 -DWANT_QT4=1 "")
-	)
+	# only used in programs
+	if ! multilib_is_native_abi ; then
+		mycmakeargs+=(
+			"CMAKE_DISABLE_FIND_PACKAGE_Fltk=ON"
```
kensington: Should these be -DCMAKE_?

a17r:
yeah, fixed now. "what was I thinking?"

shall I create another PR wrt icc_examin/oyranos cleanup?

media-libs/oyranos/oyranos-9999.ebuild
```diff
@@ -1,35 +1,35 @@
...
 
 LICENSE="BSD"
 SLOT="0"
-IUSE="X cairo cups doc exif fltk qt4 raw test"
+IUSE="X cairo cups doc exif fltk jpeg qt4 qt5 raw test tiff"
 
-RDEPEND="
+#OY_LINGUAS="cs;de;eo;eu;fr;ru" #TODO
+
+CDEPEND="
```
mgorny: Please don't call it `CDEPEND`. It's safer to reserve short *DEPEND for future variables and use something like `COMMON_DEPEND`.

media-libs/oyranos/oyranos-9999.ebuild
```diff
@@ -80,18 +88,37 @@ multilib_src_configure() {
...
 
-		$(usex cairo -DWANT_CAIRO=1 "")
-		$(usex cups -DWANT_CUPS=1 "")
-		$(usex doc -DWANT_HTML=1 "")
-		$(usex exif -DWANT_EXIV2=1 "")
-		$(usex raw -DWANT_LIBRAW=1 "")
-		$(usex X -DWANT_X11=1 "")
+	# prefer Qt5
+	if ! use qt5 ; then
+		use qt4 && mycmakeargs+=( $(cmake-utils_use_use qt4) )
```
mgorny: Hmm, so USE_QT4 only forces Qt4 over Qt5? Does it care if Qt4 is found at all? Maybe, in the common block:

```
-DUSE_QT4=$(usex '!qt5')
```

This would be simpler if it doesn't cause failures when both Qt4 & Qt5 are disabled.

media-libs/oyranos/oyranos-9999.ebuild
```diff
@@ -80,18 +88,37 @@ multilib_src_configure() {
...
+	# only used in programs
+	if ! multilib_is_native_abi ; then
+		mycmakeargs+=(
+			-DCMAKE_DISABLE_FIND_PACKAGE_Fltk=ON
+			-DCMAKE_DISABLE_FIND_PACKAGE_Qt4=ON
+			-DCMAKE_DISABLE_FIND_PACKAGE_Qt5=ON
+		)
+	else
+		mycmakeargs+=(
+			$(cmake-utils_use_find_package fltk)
```
mgorny: Again, in the common block:

```
-DCMAKE_DISABLE_FIND_PACKAGE_Fltk=$(multilib_native_usex fltk OFF ON)
```

(note the reverse logic because we won't `ON` on non-multilib)

media-libs/oyranos/oyranos-9999.ebuild
```diff
@@ -80,18 +88,37 @@ multilib_src_configure() {
 	local libdir=$(get_libdir)
 	local mycmakeargs=(
 		-DLIB_SUFFIX=${libdir#lib}
+		-DUSE_SYSTEM_ELEKTRA=YES
+		-DUSE_SYSTEM_YAJL=YES
+		$(cmake-utils_use_find_package cairo)
```
mgorny: If you follow the two other advices, I suggest dropping cmake-utils_use\* entirely in favor of less confusing/better copy-paste stuff like:

```
-DCMAKE_DISABLE_FIND_PACKAGE_CAIRO=$(usex '!cairo')
```

mgorny:
```
+  25 Jul 2015; Michał Górny <mgorny@gentoo.org>
+  -files/oyranos-9999-buildsystem.patch, metadata.xml, oyranos-9999.ebuild:
+  Update the live ebuild. https://github.com/gentoo/gentoo-portage-rsync-
+  mirror/pull/157 by a17r.
```

PR #158, opened 2015-06-29
net-libs/iojs-2.3.1: zlib fixes

jbergstroem:
..was missing because I assumed it was a required part of World.

Ref https://bugs.gentoo.org/show_bug.cgi?id=553212.


jbergstroem:
Now also includes a fix for installing `USE=bundled-libs emerge iojs` which was the result of sloppy use of `sed`. It will still fail to install because of an npm issue..

mgorny:
How about some := in deps, while already updating them? :)

jbergstroem:
Fwiw, I still haven't had time to hunt down why npm cries. To be honest, I'm pretty keen on killing `bundled-libs` anyway.

@mgorny thoughts on dropping `bundled-libs` before merging everything back to nodejs? (months away)

Edit: I'll give it one more hour; might as well..

mgorny:
Drop it!

mgorny:
Could you rebase this, please?

jbergstroem:
@mgorny we were removing `bundled-libs` in #166. Closing :)

Bonus points: if you could create a `-r1` with the changes left out in latest iojs that'd be great.

mgorny:
I barely have time to review and merge stuff :). Please create PRs with any changes you want done :).

jbergstroem:
@mgorny essentially everything in #166, as is (patrick just bumped).

PR #167, opened 2015-07-06
Version bump: net-libs/nodejs-0.12.6

jbergstroem:
Security release, more info here:
https://medium.com/@iojs/important-security-upgrades-for-node-js-and-io-js-8ac14ece5852

Also, suggesting we drop older versions of 0.12.x.


jbergstroem:
Merged to `gentoo-x86` by Patrick Lauer.

PR #168, opened 2015-07-09
dev-libs/libappindicator: enable python bindings, fix introspection use flag

anyc:
This patch:
- enables the python bindings, required for (GTK) python apps to show a systray icon in newer desktop environments like Plasma 5
- fixes the handling of the introspection use flag - now, a disabled flag disables introspection during configuration and a enabled flag requires gtk+ with introspection as well

This pull request would close these bugs:
- https://bugs.gentoo.org/show_bug.cgi?id=554078 
- https://bugs.gentoo.org/show_bug.cgi?id=512976


dev-libs/libappindicator/libappindicator-12.10.0-r200.ebuild
```diff
@@ -6,7 +6,9 @@ EAPI=5
 VALA_MIN_API_VERSION="0.16"
 VALA_USE_DEPEND="vapigen"
 
-inherit autotools eutils multilib-minimal vala
+PYTHON_COMPAT=( python2_7 )
+
+inherit autotools eutils multilib-minimal vala python-single-r1
```
mgorny: Please sort lexically whenever possible.

dev-libs/libappindicator/libappindicator-12.10.0-r200.ebuild
```diff
@@ -15,43 +17,56 @@ SRC_URI="http://launchpad.net/${PN}/${PV%.*}/${PV}/+download/${P}.tar.gz"
 LICENSE="LGPL-2.1 LGPL-3"
 SLOT="2"
 KEYWORDS="~amd64 ~x86"
-IUSE="+introspection"
+IUSE="introspection python"
```
mgorny: Any reason you no longer enable it by default?

dev-libs/libappindicator/libappindicator-12.10.0-r200.ebuild
```diff
@@ -15,43 +17,56 @@ SRC_URI="http://launchpad.net/${PN}/${PV%.*}/${PV}/+download/${P}.tar.gz"
...
 	virtual/pkgconfig[${MULTILIB_USEDEP}]
 	introspection? ( $(vala_depend) )
 "
 
-src_prepare() {
-	# Don't use -Werror
-	sed -i -e 's/ -Werror//' {src,tests}/Makefile.{am,in} || die
+multilib_src_prepare() {
+	if ! multilib_is_native_abi || ! use python; then
+		cd "${BUILD_DIR}"
```
mgorny: || die, but…

dev-libs/libappindicator/libappindicator-12.10.0-r200.ebuild
```diff
@@ -15,43 +17,56 @@ SRC_URI="http://launchpad.net/${PN}/${PV%.*}/${PV}/+download/${P}.tar.gz"
...
 
-	epatch "${FILESDIR}"/${P}-disable-python.patch
-	eautoreconf
+		epatch "${FILESDIR}"/${P}-disable-python.patch
```
mgorny: No love for conditional patching. Please prepare a proper patch adding some kind of `--enable-python`.

dev-libs/libappindicator/libappindicator-12.10.0-r200.ebuild
```diff
@@ -15,43 +17,56 @@ SRC_URI="http://launchpad.net/${PN}/${PV%.*}/${PV}/+download/${P}.tar.gz"
...
+	multilib_foreach_abi multilib_src_prepare
+}
+
 multilib_src_configure() {
+	local myconf="--disable-introspection"
+
 	if multilib_is_native_abi; then
 		local -x VALAC VAPIGEN_VAPIDIR PKG_CONFIG_PATH
 		use introspection && vala_src_prepare
+		myconf="$(use_enable introspection)"
```
mgorny: There's `multilib_native_use_enable` helper for this.

dev-libs/libappindicator/libappindicator-12.10.0-r200.ebuild
```diff
@@ -66,3 +81,7 @@ multilib_src_install_all() {
 	# installed by slot 3 as well
 	rm -r "${D}"usr/share/gtk-doc || die
 }
+
+pkg_setup() {
```
mgorny: Please order phase functions in execution order, i.e. `pkg_setup()` goes first.

anyc:
This revision should address all points.

mgorny:
That's a surprisingly good patch :). Could you submit it upstream as well, please?

mgorny:
You've missed out `REQUIRED_USE` but I'll fix that for you.

mgorny:
Hmm, since :3 variant uses introspection only… are the py switches even meaningful? Deps are certainly wrong.

anyc:
I think you're right. bindings/Makefile.am only builds python for GTK2 version. Now that I'm thinking about it, I believe, I even read somewhere that GTK3 includes a indicator mechanism which would make sense.

Hm, the launchpad site doesn't accept bugs. I guess I would have to find some Ubuntu mailinglist for this. :/

anyc:
Now, for -r300, I only included the introspection patch. I added the old disable-python patch again. It somehow looks odd to include --disable-python with the new patch although there is nothing to enable.

-r200 now also includes the REQUIRED_USE for python.

mgorny:
How about you modify the patch so that Python is always disabled when --with-gtk=3 is passed? :) Two patches for the same thing are definitely not the way to go.

anyc:
Ok, next try.

dev-libs/libappindicator/files/libappindicator-12.10.0-conditional-py-bindings.patch
```diff
@@ -0,0 +1,72 @@
...
+ PYGOBJECT_REQUIRED=0.22
+ 
+-AM_PATH_PYTHON(2.3.5)
+-AM_CHECK_PYTHON_HEADERS(,[AC_MSG_ERROR(could not find Python headers)])
++AC_ARG_ENABLE([python],
++	AC_HELP_STRING([--enable-python], [Enable python for GTK2 build]),,
++	[enable_python=yes])
+ 
+-PKG_CHECK_MODULES(APPINDICATOR_PYTHON,
++if test x"$with_gtk" = "x3" ; then
```
mgorny, line 28: Now I'm being picky but since I have other comments, you can fix this one as well. I'd suggest using consistent quoting on both sides, i.e. either `"x$with_gtk"` or `x"3"` :).

dev-libs/libappindicator/files/libappindicator-12.10.0-conditional-py-bindings.patch
```diff
@@ -0,0 +1,72 @@
...
++	enable_python=no
++fi
++
++AM_CONDITIONAL(HAS_PYTHON, test x${enable_python} != xno)
```
mgorny, line 32: Missing quoting around `${enable_python}` here.

dev-libs/libappindicator/files/libappindicator-12.10.0-conditional-py-bindings.patch
```diff
@@ -0,0 +1,72 @@
...
++	AC_MSG_RESULT($PYGTK_CODEGEN)
++fi
+ 
+ #########################
+ # Check if build tests
+@@ -278,4 +290,5 @@
+ 	Tests:         $enable_tests
+ 	Mono tests:    $have_nunit
+ 	gcov:	       $use_gcov
++	Python:        $enable_python
```
mgorny, line 71: I think printing `Python: no` for GTK+3 can be a bit confusing — users think they are missing something.

mgorny:
Sorry for the long delay. Great work, just a few more touches and upstream should love it.

mgorny:
In fact, I love this so much I'm going to try to fix the patch for you :).

mgorny:
```
  26 Jul 2015; Michał Górny <mgorny@gentoo.org>
  +files/libappindicator-12.10.0-conditional-py-bindings.patch,
  +libappindicator-12.10.0-r201.ebuild, +libappindicator-12.10.0-r301.ebuild,
  -files/libappindicator-12.10.0-disable-python.patch,
  libappindicator-12.10.0-r200.ebuild, libappindicator-12.10.0-r300.ebuild:
  Fix the build system in order to properly enable Python support. Fix
  introspection flag. https://github.com/gentoo/gentoo-portage-rsync-
  mirror/pull/168 by anyc.
```

Note that the filenames in changelog are actually wrong but whatever :P.

anyc:
Great, thank you!

PR #166, opened 2015-07-06
Version bump: net-libs/iojs-2.3.3

jbergstroem:
..also removes old versions b/c vulnerabilities.

More info here:https://medium.com/@iojs/important-security-upgrades-for-node-js-and-io-js-8ac14ece5852

Bump for nodejs shortly.

PR #172, opened 2015-07-09
version bump: net-libs/iojs-2.3.4

jbergstroem:
Supersedes #166. This upstream bump is a security update, but since we're removing support for building against a bundled openssl it doesn't affect us "as much".


mgorny:
Could you rebase, please?

mgorny:
1. No ChangeLogs in git.

PR #169, opened 2015-07-09
[dev-util/kdevelop-pg-qt] slotmove and version bump

a17r:

PR #174, opened 2015-07-12
[dev-util/kdevelop-python] Pin DEPENDs to slot 4

a17r:
dev-util/kdevplatform:4 dependency set by eclass, clean up

depends on PR #169


mgorny:
Hmm, I think KDE folks have some fancy dep generation stuff.

johu:
:+1: from kde

mgorny:
```
+  25 Jul 2015; Michał Górny <mgorny@gentoo.org> kdevelop-python-1.5.2.ebuild,
+  kdevelop-python-1.6.0.ebuild, kdevelop-python-1.7.1.ebuild:
+  Pin dependencies to slot :4, clean up redundant deps.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/174 by a17r.
```

PR #178, opened 2015-07-19
[media-gfx/exiv2] Version bump

a17r:
New ebuild using cmake. I had to hack around the fact that doc is not yet configured through cmake and still requires taglist from samples (=examples/test subdirectory).


media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+		media-gfx/graphviz
+		${PYTHON_DEPS}
+	)
+	nls? ( sys-devel/gettext )
+"
+
+DOCS=( README doc/ChangeLog doc/cmd.txt )
+
+PATCHES=(
+	"${FILESDIR}/${PN}-0.25-fix-docs.patch"
```
karolherbst, line 45: isn't a doman simplier than a patch?

media-gfx/exiv2/files/exiv2-0.25-fix-docs.patch
```diff
@@ -0,0 +1,71 @@
+--- a/CMakeLists.txt	2015-07-19 16:04:35.363628707 +0200
++++ b/CMakeLists.txt	2015-07-19 16:28:31.051680136 +0200
+@@ -216,9 +216,7 @@
+ ADD_SUBDIRECTORY( xmpsdk )
+ ADD_SUBDIRECTORY( src )
+ 
+-IF( EXIV2_ENABLE_BUILD_SAMPLES )
+-    ADD_SUBDIRECTORY( samples )
+-ENDIF( EXIV2_ENABLE_BUILD_SAMPLES )
++ADD_SUBDIRECTORY( samples )
```
karolherbst: isn't it easier to always enable EXIV2_ENABLE_BUILD_SAMPLES through cmake invocation instead of patching it here?

media-gfx/exiv2/exiv2-0.25.ebuild
a17r, line 45: simpler yes, unless I want to try and get it upstream. maybe fix other paths as well if they think GNUInstallDirs makes sense.

a17r: simpler yes, unless I want to try and get it upstream. maybe fix other paths as well if they think GNUInstallDirs makes sense.

karolherbst, line 45: yeah for upstream GNUInstallDirs makes much more sense

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	local mycmakeargs=(
+		-DEXIV2_ENABLE_BUILD_PO=YES
+		-DEXIV2_ENABLE_BUILD_SAMPLES=NO
+		-DEXIV2_ENABLE_CURL=$(usex webready)
+		-DEXIV2_ENABLE_NLS=$(usex nls)
+		-DEXIV2_ENABLE_PNG=$(usex png)
+		-DEXIV2_ENABLE_SHARED=$(usex "!static-libs")
+		-DEXIV2_ENABLE_SSH=$(usex webready)
+		-DEXIV2_ENABLE_WEBREADY=$(usex webready)
+		-DEXIV2_ENABLE_XMP=$(usex xmp)
```
maekke, line 92: those `usex` calls look just wrong. I'd guess using `cmake-utils_use_enable` would be appropriate here.

a17r, line 92: I'd guess so too, but on another occasion it was suggested to use usex instead of the various cmake-utils_use\* functions...

mgorny, line 92: Nope, they are fine. Unlike unreadable cmake-utils\* crap.

mgorny, line 92: Not to mention there's no cmake-utils_use_exiv2_enable...

a17r, line 92: ah yes. been a few days since I last looked at it.

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+
+pkg_setup() {
+	use doc && python-any-r1_pkg_setup
+}
+
+src_prepare() {
+	rm -rf msvc* build || die "Failed to remove msvc dirs"
+
+	if [[ -d po ]] ; then
+		pushd po > /dev/null
```
mgorny: ||die!

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+		for lang in *po; do
```
mgorny: I think it would be more appropriate to use `*.po` here, esp. that you're relying on `.po` below.

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+			if ! has ${lang/.po} ${LINGUAS} ; then
```
mgorny: `${lang%.po}` to strip suffix. `/` unnecessarily scans the whole string, and if someone modifies the code later some random `.po` may occur (i.e. due to EPREFIX or something :)).

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+				rm -rf ${lang}
```
mgorny: || die

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+			fi
+		done
+		popd > /dev/null
```
mgorny: ||die

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	fi
+
+	# convert docs to UTF-8
+	local i
+	for i in doc/cmd.txt; do
+		einfo "Converting "${i}" to UTF-8"
+		iconv -f LATIN1 -t UTF-8 "${i}" > "${i}~" && mv -f "${i}~" "${i}" || rm -f "${i}~"
```
mgorny: 1. Please don't keep this inline, it's ugly to read, and the `&&` `||` logic is confusing.
2. I think we really want to `die` if something fails here.
3. I'd use `.tmp` rather than `~`. `~` suggests backup, not new version.

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+	done
+
+	if use doc; then
+		einfo "Updating doxygen config"
+		doxygen 2>&1 >/dev/null -u config/Doxyfile
```
mgorny: &>/dev/null for bash. And don't we want to ||die?

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+		use doc && emake -j1 doc
+	fi
+}
+
+multilib_src_install_all() {
+	einstalldocs
+	prune_libtool_files --all
+
+	use xmp && dodoc doc/{COPYING-XMPSDK,README-XMP,cmdxmp.txt}
+	use doc && dohtml -r "${S}"/doc/html/.
```
mgorny: dohtml is deprecated, please check if `dodoc -r "${S}"/doc/html` would give the same result.

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+
+	if use examples; then
+		insinto /usr/share/doc/${PF}/examples
+		docompress -x /usr/share/doc/${PF}/examples
+		doins samples/*.cpp
```
mgorny:
```
docinto examples
dodoc samples/*.cpp
```

No point in using `doins` for docdir-files in EAPI 5+.

mgorny: For an explanation: people used to use `doins` because it implicitly disabled compression in earlier EAPIs but it no longer does in EAPI 5+.

a17r:
thx for your thorough review!

media-gfx/exiv2/exiv2-0.25.ebuild
```diff
@@ -0,0 +1,117 @@
...
+
+multilib_src_install_all() {
+	einstalldocs
+	prune_libtool_files --all
+
+	use xmp && dodoc doc/{COPYING-XMPSDK,README-XMP,cmdxmp.txt}
+	use doc && dodoc -r "${S}"/doc/html
+
+	if use examples; then
+		docinto /usr/share/doc/${PF}/examples
```
mgorny: `docinto` takes relative path, i.e. `docinto examples`

a17r: indeed, works both ways, fixed now

mgorny:
```
Files matching a file type that is not allowed:
   usr/lib/libexiv2.so.14.0.0
 * ERROR: media-gfx/exiv2-0.25::gentoo-cvs failed:
 *   multilib-strict check failed!
```

mgorny:
In other words, you need to fix their CMakeLists to support overriding `lib` with correct libdir. I think people usually introduce `CMAKE_LIBDIR_SUFFIX` variable for that.

mgorny:
@a17r, are you still there?

mgorny:
1. No ChangeLogs in git.

karolherbst:
@mgorny thanks for the info

PR #179, opened 2015-07-22
[kde-apps/libkdegames] Require openal; drop phonon

a17r:
- Old ebuild lacks DEPEND on media-libs/phonon
- Upstream dropped phonon per commit 5a13ecbb8a0e51159fa15513dc459a0d63da974c as "phonon was giving too many headaches"


kde-apps/libkdegames/libkdegames-4.14.3-r1.ebuild
```diff
@@ -0,0 +1,21 @@
...
+KEYWORDS="~amd64 ~arm ~ppc ~ppc64 ~x86 ~amd64-linux ~x86-linux"
+IUSE="debug"
+
+DEPEND="
+	media-libs/libsndfile
+	media-libs/openal
+"
+RDEPEND="${DEPEND}"
+
+PATCHES=( "${FILESDIR}/${P}-drop-phonon.patch" )
```
mgorny: Wouldn't it be simpler to just disable phonon and force openal?

mgorny: I mean via cmake rather than patching it all out.

a17r: There's no cmake switch for phonon as far as I could see. It would be simpler to just bring kde-apps/libkdegames-14.12.3 to tree but it is unclear at that point if that is going to happen (already deleted in overlay).

mgorny: Hmm but judging by the code, it uses phonon _only_ if openal&sndfile are not available, isn't that correct?

a17r: Actually, you're right. That would only leave possible silent misconfiguration with some future openal/sndfile version that might break checks, and possibly selecting phonon if someone uses --nodeps, if the patch is not applied.

mgorny: Yes. Then I think it's better to drop the large patch, esp. that we can expect the fix in newer versions. The common Gentoo idea is that we fix stuff like this for upstream, but in ebuilds we only worry about the env set by PM.

a17r: Well, the future version is probably not going to make it into portage, and development on kde4-based libkdegames has stopped, though for at least 1/2 year it is going to be in use from part of new kde-games releases.

mgorny:
```
+*libkdegames-4.14.3-r1 (26 Jul 2015)
+
+  26 Jul 2015; Michał Górny <mgorny@gentoo.org> +libkdegames-4.14.3-r1.ebuild:
+  Disable broken phonon fallback and force OpenAL, following upstream changes.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/179 by a17r.
```

PR #180, opened 2015-07-23
sys-freebsd/freebsd-sources-10.1-r5: fix FreeBSD-SA-15:13.tcp.

nigoro:
Fix Security Advisory
https://www.freebsd.org/security/advisories/FreeBSD-SA-15:13.tcp.asc


mgorny:
```
+*freebsd-sources-10.1-r5 (25 Jul 2015)
+
+  25 Jul 2015; Michał Górny <mgorny@gentoo.org> +freebsd-sources-10.1-r5.ebuild,
+  -freebsd-sources-10.1-r4.ebuild:
+  Fix FreeBSD-SA-15:13.tcp
```

PR #185, opened 2015-07-26
[dev-libs/libnatspec] Add missing pkg_setup

a17r:
Ah, I'm not the first one to hit this, see also https://bugs.gentoo.org/show_bug.cgi?id=546212


mgorny:
```
+  26 Jul 2015; Michał Górny <mgorny@gentoo.org> libnatspec-0.2.6-r1.ebuild:
+  Add proper pkg_setup() to fix USE=-python, bug #546212.
+  https://github.com/gentoo/gentoo-portage-rsync-mirror/pull/185 by a17r.
```

PR #192, opened 2015-08-06
freebsd-ubin-10.1-r2: Fix FreeBSD-SA-15:18.bsdpatch.

nigoro:
Fix FreeBSD-SA-15:18.bsdpatch.
https://www.freebsd.org/security/advisories/FreeBSD-SA-15:18.bsdpatch.asc


mgorny:
Merged.

PR #193, opened 2015-08-06
freebsd-sbin-10.1-r1: Fix FreeBSD-SA-15:19.routed.

nigoro:
Fix FreeBSD-SA-15:19.routed.
https://www.freebsd.org/security/advisories/FreeBSD-SA-15:19.routed.asc


mgorny:
Merged.

